PR #1, opened 2018-04-21
generate options types from fdb.options

yjh0502:
- I used `error-chain` for error handling because I'm familiar with it. It might better to change it to `failure`, as other sub-crates do.
 - I embedded `fdb.options` file into source. Please let me know if you have any preferred option.
 - `FdbScope::ty` and `FdbOption::ty` return `String` instead of taking `Write` (for maybe `Formatter`?) as a mutable argument. The code looks dirty but it works :(
 - [fdb_network_set_options](https://apple.github.io/foundationdb/api-c.html#c.fdb_network_set_option) and similar APIs take length and pointer as arguments, so next step might be generating `apply()` function which call underlying C API for each enum types.


bluejekyll:
This is awesome! I haven't even got a license up yet :) 

I was planning an Apache 2.0/MIT if that's ok with you, I'll add that to repo then we can merge that into this PR so that we have all that squared away.

----

- On `error-chain`: I just pushed an FdbError type last night, it's based on `Failure` which is going to replace `error-chain` in most use-cases in Rust. I think you're generator can skip generating errors and instead just use the hand-written type? https://github.com/bluejekyll/foundationdb-rs/blob/master/foundationdb/src/error.rs

- `fdb.options`: I have no opinion at-the-moment, we can clean this up later... I don't think this is a bad option, but we probably want to track a version somehow.

- "The code looks dirty but it works": All code starts out dirty, then we learn what we want and make it better ;) your decisions here seem fine to me. 

- `fdb_network_set_options`: I almost started working on the wrapper API for this last night. We'll have to see what happens with that.

----

Thanks, again for the PR. Let me get the LICENSE squared away, and then address the `FdbError` question and then let's go from there.

foundationdb-gen/src/bin/foundationdb-options-gen.rs
```diff
@@ -0,0 +1,332 @@
+extern crate inflector;
+extern crate xml;
+#[macro_use]
+extern crate error_chain;
```
bluejekyll: Failure is quite nice and simple. For the generator, you can get away with just using `format_err!` and returning the `failure::Error` type. in the result type, example:

```rust
fn example() -> Result<Type, failure::Error> {
   Err(format_err!("my failure message: {}", extra_info)
}
```

Can we switch to that instead? I believe most of Rust is going to be going in this direction.

foundationdb/src/options.rs
```diff
@@ -0,0 +1,615 @@
...
+impl ConflictRangeType {
+    pub fn code(&self) -> fdb::FDBConflictRangeType {
+        match *self {
+            ConflictRangeType::Read => 0,
+            ConflictRangeType::Write => 1,
+        }
+    }
+}
+
+pub enum ErrorPredicate {
```
bluejekyll: Can you look at `error::FdbError` in the `foundationdb` crate? We should maybe remove my hand-rolled version.

foundationdb/src/options.rs
```diff
@@ -0,0 +1,615 @@
...
+    /// Returns ``true`` if the error indicates the transaction may have succeeded, though not in a way the system can verify.
+    MaybeCommitted,
+    /// Returns ``true`` if the error indicates the transaction has not committed, though in a way that can be retried.
+    RetryableNotCommitted,
+}
+
+impl ErrorPredicate {
+    pub fn code(&self) -> fdb::FDBErrorPredicate {
+        match *self {
+            ErrorPredicate::Retryable => 50000,
```
bluejekyll: Technically these values are `fdb::FDBErrorPredicate_FDB_ERROR_PREDICATE_MAYBE_COMMITTED`, rather than the raw value. Not sure if this is something the generator could write instead?

yjh0502:
Thanks for the feedback!

 - I'm happy with Apache 2.0/MIT
 - For error types, I just `error-chain` mainly for `ensure!`/`bail!` macros which helps debugging during iteration. I'll replace it into hand-written error types.

For `fdb_network_set_options` things, I added a `unsafe fn apply()` which is a helper functions. You can use those functions for initial prototyping :)

 - https://github.com/bluejekyll/foundationdb-rs/pull/1/commits/53c9f455db8305439499029a07b50d4ef80d357c#diff-cd3b1f98c824b08ddc1d3579abe50a0dR371
 - https://github.com/bluejekyll/foundationdb-rs/pull/1/commits/53c9f455db8305439499029a07b50d4ef80d357c#diff-cd3b1f98c824b08ddc1d3579abe50a0dR130

foundationdb-gen/src/bin/foundationdb-options-gen.rs
yjh0502: I changed it to `failure`.

bluejekyll:
Thanks again for the PR, if this is ready I'll merge it in.

I'm working on travis scripts right now, so we should get some testing options.

foundationdb/src/options.rs
yjh0502: I replaced `FdbErrorPredicate` with `options::ErrorPredicate`. I'm not sure about name of `fn code() ->  fdb::FDBErrorPredicate`. Its okay for other enum types which might have seperated value like `String`, bug for `ErrorPredicate` case `fn value() -> ...` sounds better.

foundationdb/src/options.rs
yjh0502: I just tried to change those values to generated one, but there are some mismatches between names on `fdb.options` and names on `fdb_c.h`. Here's a list.

https://gist.github.com/yjh0502/cd19273e15464e5e2c377fc9f3ce3e03

 - Some of them are just singular/plural problem, for example `WATCH/WATCHES`, `BYTE/BYTES`. I can handle it with some brute-force replace rules.
 - It seems that options with `hidden=true` is not available on c header. I filtered those options.
 - There's one remaining non-regular case, `FDBMutationType_FDB_MUTATION_TYPE_SET_VERSIONSTAMPED_KEY`, which should not be converted to `..._KEYS`. I just hard-coded the case :(

bluejekyll:
Thanks again for the PR, this is a massive help!

PR #3, opened 2018-04-21
add first working example

yjh0502:
I just wrote a working foundationdb example. I submit this PR to propose a shape of internal structures and API.

 - I used `futures` 0.1 intentionally. `futures` 0.1 has enough ecosystem to build some working apps, for example database drivers and `hyper`, while there's not that much libraries for `futures` 0.2. I believe the client should built on top of `futures` 0.1 at this time.
 - `FdbFuture` is a wrapper for `FDBFuture`, but with rust `Futures` crate. There's an heap allocation (`Box<Task>`) which is hard to avoid because `foundationdb` is completion-based and rust `Futures` is poll-based.
 - `FdbFutureResult` is a result of a future, which contains a pointer to completed `FDBFuture`. It can be resolved into actual result like `FDBCluster`, `FDBDatabase`, or some others.
 - ~There should be `FdbTrxFuture` like things to retain reference count to the database handle while running query. In current example, dropping `FdbDatabase` while querying will clean-up all resources before transaction finishes.~ I added `FdbClusterGet`/`FdbTrxGet` to fix resource leak/use after free.

Misc.
 - It seems that there's no way to mix network thread of `foundationdb` with other I/O loop.
 - There are tons of `Arc`s to manage internal resources. It might hurt performance, but I can't find better solution for now.

### DIFF
--- a/foundationdb/examples/hello.rs
+++ b/foundationdb/examples/hello.rs
@@ -1,0 +1,293 @@
+extern crate foundationdb;
+extern crate foundationdb_sys;
+extern crate futures;
+extern crate tokio_core;
+
+use std::sync::Arc;
+
+use foundationdb::error;
+use foundationdb_sys as fdb;
+
+use futures::future::*;
+
+use error::FdbError;
+
+type Result<T> = std::result::Result<T, FdbError>;
+
+#[derive(Clone)]
+struct FdbCluster {
+    inner: Arc<FdbClusterInner>,
+}
+impl FdbCluster {
+    fn new(inner: *mut fdb::FDBCluster) -> Self {
+        Self {
+            inner: Arc::new(FdbClusterInner::new(inner)),
+        }
+    }
+
+    //TODO: impl Future
+    fn create_database(&self) -> Box<Future<Item = FdbDatabase, Error = FdbError>> {
+        let f = unsafe {
+            let f_db = fdb::fdb_cluster_create_database(self.inner.inner, b"DB" as *const _, 2);
+            let cluster = self.clone();
+            FdbFuture::new(f_db)
+                .and_then(|f| f.get_database())
+                .map(|db| FdbDatabase {
+                    cluster,
+                    inner: Arc::new(FdbDatabaseInner::new(db)),
+                })
+        };
+        Box::new(f)
+    }
+}
+
+struct FdbClusterInner {
+    inner: *mut fdb::FDBCluster,
+}
+impl FdbClusterInner {
+    fn new(inner: *mut fdb::FDBCluster) -> Self {
+        Self { inner }
+    }
+}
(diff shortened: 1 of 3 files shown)


foundationdb/examples/hello.rs
```diff
@@ -0,0 +1,428 @@
...
+        }
+
+        let g = FdbFutureResult::new(self.f);
+        self.f = std::ptr::null_mut();
+
+        Ok(Async::Ready(g))
+    }
+}
+
+extern "C" fn fdb_future_callback(
```
bluejekyll: nice work, this looks good.

foundationdb/examples/hello.rs
```diff
@@ -0,0 +1,428 @@
...
+    let mut core = tokio_core::reactor::Core::new().unwrap();
+
+    let handle = unsafe {
+        let version = fdb::fdb_get_max_api_version();
+        let err = fdb::fdb_select_api_version_impl(version, version);
+        if err != 0 {
+            panic!("fdb_select_api_version: {:?}", FdbError::from(err));
+        }
+
+        let err = fdb::fdb_setup_network();
```
bluejekyll: Do you have any thoughts on #4 ? I want to try and make some safety guarantees around initialization. Everything in FoundationDB seems to be static, which is annoying.

bluejekyll:
I'm going to merge this in. I agree that the FoudationDB API has some disappointing limitations. Their crazy use of statics is definitely not great. Over time, perhaps we can reimplement that C API directly in Rust...

foundationdb/examples/hello.rs
yjh0502: How about benchmarking `env_logger`? For me `env_logger` looks quite similar with `Network`.

 - Both `env_logger` / `Network` handles global static resources, and both could be initialized only once, with only one config.
 - `env_logger` should not be initialized by libraries, and I believe it also applies to `Network` too.

For now, I thinks it's better to expose those APIs with static functions, like `env_logger::{init/try_init}`.

```rust
use foundationdb as fdb;
fn main() {
  fdb::init();
  let handle = fdb::start_network();
  // do something
  fdb::stop_network();
  handle.join().expect("failed to join");
}
```

If user wants to customize a network config, they can use `Builder`, just like `env_logger::Builder`.

```rust
use foundationdb as fdb;
fn main() {
  let builder = fdb::Builder().set_network_version(500).set_option(fdb::options::NetworkOption::TraceEnable(true));
  builder.init(); // panic if called twice, just like `env_logger::Builder::init()`
  let handle = fdb::start_network();
  // do something
  fdb::stop_network();
  handle.join().expect("failed to join");
}
```

It seems that calling `fdb_run_network` multiple times causes segfault in library. I think we can fix the problem by having global state of whether the network is running or not.

PR #4, opened 2018-04-21
static Network related functions

bluejekyll:


bluejekyll:
FYI @yjh0502 there are now two builders, FdbApiBuilder and NetworkBuilder. The NetworkBuilder requires the FdbApi output, which can only be created form the api builder.

I'll create some higher level interfaces in the root of the library, like env_logger. Still need to add a check for run only being called once. I'm planning on an atomic_bool for that...

bluejekyll:
Ok, I think this is ready. For a try_init we'll probably need to track if the Api has been initialized as well.

PR #7, opened 2018-04-22
move examples code to lib

yjh0502:
Move codes from `examples` to lib.

 - naming: I used `FdbFuture`/`FdbFutureResult` because it sounds similar with `FdbError`. For other things, I used name without `Fdb` prefix, like `Cluster`, `Transaction`, ...
 - namespace: At this time I split each component into a file. Some of them looks too small to be a separated file, like `Cluster`
 - `Result`: I added `error::Result<T> = std::result::Result<T, error::FdbError>` for convenience.

### DIFF
--- a/foundationdb/examples/hello.rs
+++ b/foundationdb/examples/hello.rs
@@ -38,8 +38,17 @@ fn example_set_get() -> Box<Future<Item = (), Error = FdbError>> {
     Box::new(fut)
 }
 
+#[cfg(target_os = "linux")]
+fn default_config_path() -> &'static str {
+    "/etc/foundationdb/fdb.cluster"
+}
+#[cfg(target_os = "macos")]
+fn default_config_path() -> &'static str {
+    "/usr/local/etc/foundationdb/fdb.cluster"
+}
+
 fn example_get_multi() -> Box<Future<Item = (), Error = FdbError>> {
-    let fut = Cluster::new("/etc/foundationdb/fdb.cluster")
+    let fut = Cluster::new(default_config_path())
         .and_then(|cluster| cluster.create_database())
         .and_then(|db| result(db.create_trx()))
         .and_then(|trx| {
--- a/foundationdb/examples/hello.rs
+++ b/foundationdb/examples/hello.rs
@@ -18,7 +18,7 @@ fn example_set_get() -> Box<Future<Item = (), Error = FdbError>> {
             trx.set(b"hello", b"world");
             trx.commit()
         })
-        .and_then(|db| result(db.create_trx()))
+        .and_then(|trx| result(trx.database().create_trx()))
         .and_then(|trx| trx.get(b"hello"))
         .and_then(|res| {
             let val = res.value();
@@ -28,7 +28,7 @@ fn example_set_get() -> Box<Future<Item = (), Error = FdbError>> {
             trx.clear(b"hello");
             trx.commit()
         })
-        .and_then(|db| result(db.create_trx()))
+        .and_then(|trx| result(trx.database().create_trx()))
         .and_then(|trx| trx.get(b"hello"))
         .and_then(|res| {
             eprintln!("value: {:?}", res.value());
--- a/foundationdb/examples/hello.rs
+++ b/foundationdb/examples/hello.rs
@@ -2,352 +2,16 @@ extern crate foundationdb;
 extern crate foundationdb_sys;
 extern crate futures;
 
-use std::sync::Arc;
-
-use foundationdb::error;
-use foundationdb_sys as fdb;
+use foundationdb::*;
+use foundationdb_sys as fdb_sys;
 
 use futures::future::*;
-use futures::Async;
 
 use error::FdbError;
 
-type Result<T> = std::result::Result<T, FdbError>;
-
-#[derive(Clone)]
-struct FdbCluster {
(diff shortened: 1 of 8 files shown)


foundationdb/src/future.rs
```diff
@@ -0,0 +1,117 @@
...
+        if ready == 0 {
+            return Ok(Async::NotReady);
+        }
+
+        let err = unsafe { fdb::fdb_future_get_error(self.f) };
+        if err != 0 {
+            return Err(FdbError::from(err));
+        }
+
+        let g = FdbFutureResult::new(self.f);
```
bluejekyll: let's add a comment here that FdbFutureResult is taking ownership of the FDBFuture

foundationdb/src/future.rs
```diff
@@ -0,0 +1,117 @@
...
+        self.f = std::ptr::null_mut();
+
+        Ok(Async::Ready(g))
+    }
+}
+
+extern "C" fn fdb_future_callback(
```
bluejekyll: let's put a comment here, that this may be called across different threads (according to the docs)

foundationdb/src/future.rs
```diff
@@ -0,0 +1,117 @@
...
+
+pub struct FdbFutureResult {
+    f: *mut fdb::FDBFuture,
+}
+impl FdbFutureResult {
+    pub(crate) fn new(f: *mut fdb::FDBFuture) -> Self {
+        Self { f }
+    }
+
+    pub(crate) unsafe fn get_cluster(&self) -> Result<*mut fdb::FDBCluster> {
```
bluejekyll: Any reason not to internally convert to a Cluster (same question for the other interfaces) and return that instead? Then this could be a safe function.

foundationdb/src/future.rs
```diff
@@ -0,0 +1,117 @@
...
+                &mut out_len as *mut _,
+            )
+        };
+        if err != 0 {
+            return Err(FdbError::from(err));
+        }
+        if present == 0 {
+            return Ok(None);
+        }
+        let slice = unsafe { std::slice::from_raw_parts(out_value, out_len as usize) };
```
bluejekyll: nice job. This slice is borrowed to the lifetime of the future, and the backing memory is released on `fdb_future_destroy` in the Drop impl. Can we put a comment here declaring that? I had to double check all the docs.

foundationdb/src/transaction.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+}
+
+pub struct TrxCommit {
+    trx: Transaction,
+    inner: FdbFuture,
+}
+impl Future for TrxCommit {
+    //should be ()?
+    type Item = Database;
```
bluejekyll: is this to make chaining of operations easier?

foundationdb/examples/hello.rs
```diff
@@ -2,352 +2,16 @@ extern crate foundationdb;
...
-) {
-    let task: *const futures::task::Task = callback_parameter as *const _;
-    let task: &futures::task::Task = unsafe { std::mem::transmute(task) };
-    task.notify();
-}
-
 //TODO: impl Future
 fn example_set_get() -> Box<Future<Item = (), Error = FdbError>> {
-    let fut = FdbCluster::new("/etc/foundationdb/fdb.cluster")
+    let fut = Cluster::new("/etc/foundationdb/fdb.cluster")
```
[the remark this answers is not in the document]
yjh0502: I added `#[cfg(target_os = ..)]` to determine default config path.

foundationdb/src/cluster.rs
```diff
@@ -0,0 +1,71 @@
...
+use futures::{Async, Future};
+use std;
+use std::sync::Arc;
+
+use database::*;
+use error::*;
+
+#[derive(Clone)]
+pub struct Cluster {
+    inner: Arc<ClusterInner>,
```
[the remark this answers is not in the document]
yjh0502: I added a TODO comment. It seems that cluster is thread-safe according to Golang API documentation and source, but there's no harm to double-check it.

https://godoc.org/github.com/apple/foundationdb/bindings/go/src/fdb#Cluster
https://github.com/apple/foundationdb/blob/master/bindings/go/src/fdb/cluster.go#L50

foundationdb/src/future.rs
yjh0502: I added a comment.

foundationdb/src/future.rs
yjh0502: I added a comment.

foundationdb/src/future.rs
yjh0502: In some case, those function cannot return safe type directly. For example, `FdbFutureResult::get_database` cannot return `Database` directly because `Database` should retain reference count to `Cluster` via `Arc<Cluster>`.

foundationdb/src/future.rs
yjh0502: I added a comment.

foundationdb/src/transaction.rs
yjh0502: That's my original intention, however I changed it to `Transaction` because there's an transaction API which could be called after commit, `fdb_transaction_get_committed_version`. That's my mistake :(

bluejekyll:
Excellent work. I’ll merge this in.

PR #23, opened 2018-04-23
update bindings from linux bindgen

yjh0502:
@bluejekyll I checked that this version works on macOS. Could you double-check if it works on your environment?


bluejekyll:
yeah, just checked. Looks good. Was that hand-edited? This will be annoying to upgrade if that's the case.

yjh0502:
It is not hand-edited and generated by `bindgen` on linux. I think we can go with this version.

Issue #19, opened 2018-04-23
missing API: watch


yjh0502:
https://apple.github.io/foundationdb/api-c.html#c.fdb_transaction_watch

PR #25, opened 2018-04-23
add `Transaction::watch`, #19

yjh0502:


foundationdb/examples/watch.rs
```diff
@@ -0,0 +1,120 @@
...
+/// generate random string. Foundationdb watch only fires when value changed, so updating with same
+/// value twice will not fire watches. To make examples work over multiple run, we use random
+/// string as a value.
+fn random_str(len: usize) -> String {
+    use rand::Rng;
+    let mut rng = rand::thread_rng();
+    rng.gen_ascii_chars().take(len).collect::<String>()
+}
+
+fn example_watch() -> Box<Future<Item = (), Error = FdbError>> {
```
bluejekyll: instead of examples, what about moving this to the a new tests/ directory, then it could be an integration test running there...

foundationdb/examples/watch.rs
```diff
@@ -0,0 +1,120 @@
...
+        let error = network.run();
+
+        if let Err(error) = error {
+            panic!("fdb_run_network: {}", error);
+        }
+    });
+
+    network.wait();
+
+    example_watch().wait().expect("failed to run");
```
bluejekyll: at some point we should add an integration test that would validate this working inside the context of a tokio reactor. The watch future could be "spawned" into the reactor, and then be used to trigger other events there...

foundationdb/examples/watch.rs
yjh0502: I moved it into `tests`.

Issue #15, opened 2018-04-23
missing API: atomic_op


yjh0502:
https://apple.github.io/foundationdb/api-c.html#c.fdb_transaction_atomic_op

PR #26, opened 2018-04-23
add `Transaction::atomic_op`, #15

yjh0502:
I update `options.rs` to remove `Vec<u8>` from `MutationType` enum. It breaks API but I'm not sure if we should bump version number, as those options are not used in API before.

It seems that Golang API splits all operations into seperated functions [1], so it has `BitAnd`, `BitOr` API functions. I think taking enum as a parameter is better than having functions for each enums.

I think current function signature is okay for all `MutationType` except `Add`. For `Add`, Foundationdb requires little-endian two's complement encoding, so caller should encode value with `byteorder` or `std::mem::transmute`. We might better to add helper API like `atomic_add_i64(&self, key: &[u8], value: i64)` for those cases.

[1] https://godoc.org/github.com/apple/foundationdb/bindings/go/src/fdb#Transaction

### DIFF
--- a/foundationdb/tests/atomic.rs
+++ b/foundationdb/tests/atomic.rs
@@ -1,0 +1,118 @@
+// Copyright 2018 foundationdb-rs developers, https://github.com/bluejekyll/foundationdb-rs/graphs/contributors
+//
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+extern crate byteorder;
+extern crate foundationdb;
+extern crate foundationdb_sys;
+extern crate futures;
+
+use foundationdb::*;
+
+use byteorder::ByteOrder;
+use futures::future::*;
+
+use error::FdbError;
+
+//TODO: impl Future
+fn atomic_add(db: Database, key: &[u8], value: i64) -> Box<Future<Item = (), Error = FdbError>> {
+    let trx = match db.create_trx() {
+        Ok(trx) => trx,
+        Err(e) => return Box::new(err(e)),
+    };
+
+    let val = {
+        let mut buf = [0u8; 8];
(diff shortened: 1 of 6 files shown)


bluejekyll:
> I update options.rs to remove Vec<u8> from MutationType enum. It breaks API but I'm not sure if we should bump version number, as those options are not used in API before.

I don't think we should worry too much about breaking vs. non-breaking. People should expect that a lot of changes are coming, so we can just bump to 0.2.0...

> It seems that Golang API splits all operations into seperated functions [1], so it has BitAnd, BitOr API functions. I think taking enum as a parameter is better than having functions for each enums.

I agree, Go doesn't have as nice a type system as Rust for things like this, so we should definitely leverage Rust's enums for this.

> I think current function signature is okay for all MutationType except Add. For Add, Foundationdb requires little-endian two's complement encoding, so caller should encode value with byteorder or std::mem::transmute. We might better to add helper API like atomic_add_i64(&self, key: &[u8], value: i64) for those cases.

I think it would be nice to abstract away any platform conversion code, we can always expose it later if people want direct access.

foundationdb/tests/atomic.rs
```diff
@@ -0,0 +1,118 @@
...
+
+use foundationdb::*;
+
+use byteorder::ByteOrder;
+use futures::future::*;
+
+use error::FdbError;
+
+//TODO: impl Future
+fn atomic_add(db: Database, key: &[u8], value: i64) -> Box<Future<Item = (), Error = FdbError>> {
```
bluejekyll: is there a big benefit for exposing individual Atomic operations, as opposed to a generic interface that calls through to atomic_op?

yjh0502: I wrote it to simplify main test flow. For `Transaction::atomic_add`, I think it might worth to have because without it users should add `byteorder` as a dependency and write severeal lines of code for one atomic add operations, as this `atomic_add` does [1].

[1] https://github.com/bluejekyll/foundationdb-rs/pull/26/files/c400b23781408214ce8248c2320d47e21a05be3f#diff-54a7af36c834a0fb789fab78a0dce2f6R27

yjh0502:
I just rebased the PR. The main issue in the PR was `MutationType`, so it could be merged after code review.

PR #28, opened 2018-04-23
add `Transaction::get_range`, #14

yjh0502:
I encountered tons of design issues which I don't have an answer. I need some help :). I tried to write `TODO` comment for each problems. Here's some other issues which is not written in `TODO`.

 - `KeyValuesStream`: It returns a chunk of `KeyValue<'a>` at a time, instead of emitting a key-value at a time. I thought that it's better to expose lowest-level API without introducing some overhead, but it might introduce some usability issue.
 - About `Transaction::get_range`
   - I implemented it to return a stream, but it gives a illusion to users that they might be able to scan very large range of key-values in single transaction, while foundationdb only allows single transaction with maximum 5 seconds (and they recommends much shorter transaction duration for performance).
   - In current design, there's no easy way to re-start new transaction to continue scan. We might change API as following to allow user to split single scan over multiple transactions, so that can use `Future::loop_fn` to scan range of keys.
```rust
impl Transaction { fn get_range(&self, opt: RangeOption) -> TrxRange; }
impl Future for TrxRange { type Item = RangeResult; type Error = FdbError; }
impl RangeResult {
  fn keyvalues(&self) -> Result<KeyValues<'a>>;
  fn next(&self) -> Option<RangeOption>; // caller can pass a return value for next `get_range` call
}
```
 - `KeySelector<'a>`/`OwnedKeySelector`: I'm not sure how to clean up `keyselector`. Maybe we can just remove `KeySelector<'a>`, but I'm feeling guilty for copying values... :(

### DIFF
--- a/foundationdb/examples/range.rs
+++ b/foundationdb/examples/range.rs
@@ -1,0 +1,90 @@
+// Copyright 2018 foundationdb-rs developers, https://github.com/bluejekyll/foundationdb-rs/graphs/contributors
+//
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+extern crate foundationdb;
+extern crate foundationdb_sys;
+extern crate futures;
+extern crate rand;
+
+use foundationdb::keyselector::*;
+use foundationdb::*;
+
+use futures::future::*;
+use futures::stream::*;
+
+use error::FdbError;
+
+fn random_str(len: usize) -> String {
+    use rand::Rng;
+    let mut rng = rand::thread_rng();
+    rng.gen_ascii_chars().take(len).collect::<String>()
+}
+
+fn example_get_range() -> Box<Future<Item = (), Error = FdbError>> {
+    const N: usize = 10000;
+
+    let fut = Cluster::new(foundationdb::default_config_path())
+        .and_then(|cluster| cluster.create_database())
+        .and_then(|db| result(db.create_trx()))
+        .and_then(|trx| {
+            let key_begin = "test-range-";
+            let key_end = "test-range.";
+
+            trx.clear_range(key_begin.as_bytes(), key_end.as_bytes());
+
+            for _ in 0..N {
+                let key = format!("{}-{}", key_begin, random_str(10));
+                let value = random_str(10);
+                trx.set(key.as_bytes(), value.as_bytes());
+            }
+
+            let begin = KeySelector::first_greater_or_equal(key_begin.as_bytes());
+            let end = KeySelector::first_greater_than(key_end.as_bytes());
+            let opt = transaction::RangeOptionBuilder::new(begin, end).build();
+
+            let stream = trx.get_range(opt);
+            stream
+                .fold(0, |count, item| {
+                    let kvs = item.keyvalues()?;
+                    Ok(count + kvs.as_ref().len())
+                })
+                .map(|count| {
+                    if count != N {
+                        panic!("count expected={}, found={}", N, count);
+                    }
+                    eprintln!("count: {:?}", count);
+                })
+        });
+
+    Box::new(fut)
+}
+
+fn main() {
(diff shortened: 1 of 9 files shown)


foundationdb/examples/range.rs
```diff
@@ -0,0 +1,90 @@
...
+
+use error::FdbError;
+
+fn random_str(len: usize) -> String {
+    use rand::Rng;
+    let mut rng = rand::thread_rng();
+    rng.gen_ascii_chars().take(len).collect::<String>()
+}
+
+fn example_get_range() -> Box<Future<Item = (), Error = FdbError>> {
```
bluejekyll: for the Boxed futures, I bet we could come up with a Generic type so that we don't need to define so many. I'll take a look at this.

foundationdb/src/transaction.rs
```diff
@@ -287,6 +509,167 @@ impl Future for TrxGet {
...
+        Ok(())
+    }
+}
+
+impl<'a> Stream for KeyValuesStream {
+    type Item = GetRangeResult;
+    type Error = FdbError;
+
+    fn poll(&mut self) -> std::result::Result<Async<Option<Self::Item>>, Self::Error> {
+        if self.inner.is_none() {
```
bluejekyll: Would it be here that we'd track the `more` flag and then trigger another query to get results after the last key received?

bluejekyll:
> I implemented it to return a stream, but it gives a illusion to users that they might be able to scan very large range of key-values in single transaction, while foundationdb only allows single transaction with maximum 5 seconds (and they recommends much shorter transaction duration for performance).

is this where the `more` flag is useful? That is, can the Stream interface be used to trigger more queries if the more flag is set?

yjh0502:
Thanks for the review! At first, this PR is not finished yet as you noticed. Sorry for confusion.

The problem is, while the Stream interface can be used to trigger more `get_range` queries when it encountered `more` flag to fetch next result (actually, this is what current implementation is doing), it could be failed before fetching all ranges when transaction ages too old.

For example, let's assume that there are billions of keys in DB and user starts `get_range` from `[0x00]` to `[0xff]` to scan all those key-values. Current implementation of `KeyValuesStream` will call `get_range` repeatedly if there's `more` flag, regardless of how old underling current transaction is. After 5 seconds, `get_range` will fail with `transaction_too_old` error [1] and `KeyValuesStream` passes this `Error` to users. At this point, it's quite hard for users to maintain necessary information to re-start next transaction to continue scanning database. We cannot automatically restart next transaction to solve the problem, because it will break consistency guarantee so user will see inconsistent set of results for single stream of result.

[1] https://apple.github.io/foundationdb/known-limitations.html#long-running-transactions

bluejekyll:
thank you for that excellent write-up of the issue. it makes better sense now.

yjh0502:
I just revised the API, and I think it's good to go for now. Here's a summary.

 - `get_range` returns `TrxGetRange` future, which will be resolved to `GetRangeResult`. A result might not contains all results specified by initial `RangeOption`. User should check it by using either `GetRangeResult::next()`, which will returns next `RangeOption` that should be used by following `get_range` call.

 - `get_ranges` returns `RangeStream` stream, which will be resolved to sequence of `GetRangeResult`. It will tries to fetch all results specified by initial `RangeOption`. If it fails in the middle (due to `transaction_too_long` or something), it returns `(RangeOption, FdbError)` as an error. A caller can check which error is happened by checking `FdbError`. If the caller wants to continue scan with new transaction, `RangeOption` could be used.

foundationdb/src/transaction.rs
```diff
@@ -318,6 +526,178 @@ impl Future for TrxGet {
...
+
+    /// Returns the values associated with this get
+    pub fn keyvalues(&self) -> KeyValues {
+        self.inner.get_keyvalue_array().unwrap()
+    }
+
+    /// Returns `None` if all results are returned, and returns `Some(_)` if there are more results
+    /// to fetch. In this case, user can fetch remaining results by calling
+    /// `Transaction::get_range` with returned `RangeOption`.
+    pub fn next(&self) -> Option<RangeOption> {
```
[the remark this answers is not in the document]
bluejekyll: Ignore this comment. I reviewed in more depth, the next is being used in stream... fully understand it now.

bluejekyll:
Excellent work! merging.

PR #32, opened 2018-04-25
move examples into tests, clean-up tests

yjh0502:
I moved all examples/tests to `test.rs`, to organize some (integration?) tests.

There are some alternatives to manage tests
 1. Keep all tests in single file, and use custom runner for each test: It works, but we cannot get usual `cargo test` style output or do test filtering for the test.
 2. Keep all tests in single file, and use `#[test]` for each test: Rust does not support setup/teardown for `#[test]`. To run multiple `#[test]` function in single binary, we should have global fdb network for those tests, and it seems that there's no easy way to clean-up the global network.
 3. Keep each tests on a seperated file, and add helper function for setup/teardown: I dislike this approach because an output of `cargo test` becomes noisy as tests grows.

I choose second option and add additional test for testing network setup/teardown only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,5 @@ script:
   - cargo test --manifest-path foundationdb-sys/Cargo.toml --all-features
   - cargo test --manifest-path foundationdb-gen/Cargo.toml --all-features
   - cargo test --manifest-path foundationdb/Cargo.toml --all-features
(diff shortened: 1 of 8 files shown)


yjh0502:
Thanks for the review. I re-organized tests as you suggested. As always, feedback is welcome!

Issue #29, opened 2018-04-24
Should retry a transaction on conflict


yjh0502:
https://apple.github.io/foundationdb/api-c.html#c.fdb_transaction_on_error

yjh0502:
Golang API provides `db.Transact` which will retry on conflict or some transient failure.
https://godoc.org/github.com/apple/foundationdb/bindings/go/src/fdb#hdr-Basic_Usage

Python API provides `@fdb.transactional` decorator.
https://apple.github.io/foundationdb/api-python.html#transactional-decoration

Java API provides `Database.run`: `default<T> T run(java.util.function.Function<? super Transaction,T> retryable)`
https://apple.github.io/foundationdb/javadoc/com/apple/foundationdb/Database.html#run-java.util.function.Function-

yjh0502:
Done

PR #34, opened 2018-04-25
Add `Database::transact`, #29

yjh0502:


foundationdb/src/database.rs
```diff
@@ -49,6 +51,45 @@ impl Database {
...
+                        // try to commit the transaction
+                        trx0.commit().map(|_| res)
+                    })
+                    .then(|res| match res {
+                        Ok(v) => {
+                            // committed
+                            Ok(Loop::Break(v))
+                        }
+                        Err(e) => {
+                            if e.should_retry() {
```
bluejekyll: Should this loop be bounded? it looks like currently it will potentially be an infinite loop in the worst case...

foundationdb/src/database.rs
```diff
@@ -49,6 +51,45 @@ impl Database {
...
+    pub fn transact<F, Fut, Item>(&self, f: F) -> Box<Future<Item = Fut::Item, Error = FdbError>>
+    where
+        F: FnMut(Transaction) -> Fut + 'static,
+        Fut: Future<Item = Item, Error = FdbError> + 'static,
+        Item: 'static,
+    {
+        let db = self.clone();
+
+        let f = result(db.create_trx()).and_then(|trx| {
+            loop_fn((trx, f), |(trx, mut f)| {
```
bluejekyll: very nice and concise implementation!

foundationdb/tests/watch.rs
```diff
@@ -71,8 +71,10 @@ fn test_watch_without_commit() {
         .and_then(|trx| {
             eprintln!("setting watch");
 
-            // trx will be dropped without `commit`, so a watch will be canceled
-            trx.watch(KEY)
+            let f = trx.watch(KEY);
+            // `trx` is canceled, so a watch will be canceled
+            trx.cancel();
```
bluejekyll: transactions are still implicitly canceled with drop, right?

foundationdb/src/error.rs
```diff
@@ -82,4 +89,17 @@ impl FdbError {
...
+    // `fdb_transaction_on_error` requires error code, so the code should accessible in the crate.
+    pub(crate) fn code(&self) -> i32 {
+        self.error_code
+    }
+
+    pub(crate) fn set_should_retry(&mut self, should_retry: bool) {
+        self.should_retry = should_retry;
+    }
+
+    pub(crate) fn should_retry(&self) -> bool {
```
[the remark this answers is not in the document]
yjh0502: The documentation says to restart the application code if `FDBFuture` returns by `fdb_transaction_on_error` is not an error. We set `should_retry` flag only when `TrxErrFuture` returns `Ok(_)`, so I think it's okay.

https://apple.github.io/foundationdb/api-c.html#c.fdb_transaction_on_error

foundationdb/src/database.rs
yjh0502: You're right. As an example, if we remove `RetryLimit` from `test_transact` it will run indefinitely. I believe it's okay to leave it as is because the behavior matches with other implementations[1].

[1] https://github.com/apple/foundationdb/blob/master/bindings/go/src/fdb/database.go#L86-L104

foundationdb/tests/watch.rs
yjh0502: The API behavior is changed as you noticed. Now transaction future is handled by `TrxFuture` not `FdbFuture`, and `TrxFuture` maintains reference count for `Transaction` because pointer to the transaction is needed to check backoff behavior on error (more specifically, `fdb_transaction_on_error` requires `*mut fdb::FDBTransaction`).

A return value of `Transaction::watch` is a `TrxWatch` which is a future of type `Future<Item=(), Error=FdbError>`. It consists of `FdbFuture`, so the transaction is not aborted as long as `TrxWatch` alives.

We can revert the behavior, but it will break `Database::transact` if `Transaction::watch` returns retryable error. Maybe we should check if `fdb_transaction_watch` can return retryble error.

foundationdb/src/database.rs
bluejekyll: Can we add a comment documenting the reliance on the RetryLimit as the upper bound for this? That way people won't accidentally be caught off-guard by it.

foundationdb/tests/watch.rs
bluejekyll: My biggest concern is that the transaction is canceled in the drop, as that seems to guard against any dangling transactions. I think I'm understanding what your concern is, and you are in a better position to make the call here. But if I have a preference, it's that we try not to have an interface where cancel needs to be called explicitly.

foundationdb/src/database.rs
yjh0502: Done.

foundationdb/tests/watch.rs
yjh0502: I understood and agree. I reverted this change.

bluejekyll:
Great job! Merging.

Issue #20, opened 2018-04-23
missing API: transaction misc


yjh0502:
fdb_transaction_add_conflict_range
fdb_transaction_get_read_version
fdb_transaction_get_versionstamp
fdb_transaction_set_read_version

PR #38, opened 2018-04-27
trx: versionstamp, {get,set}_read_version, #20

yjh0502:


foundationdb/src/transaction.rs
```diff
@@ -754,6 +799,68 @@ impl Future for TrxWatch {
     }
 }
 
+/// A versionstamp is a 10 byte, unique, monotonically (but not sequentially) increasing value for
+/// each committed transaction. The first 8 bytes are the committed version of the database. The
+/// last 2 bytes are monotonic in the serialization order for transactions.
+#[derive(Clone, Copy)]
+pub struct Versionstamp([u8; 10]);
```
bluejekyll: I’m confused about this size, is it a context thing? The java client claims this is 12: https://apple.github.io/foundationdb/javadoc/index.html?com/apple/foundationdb/tuple/Versionstamp.html

Is it a context thing? Where in the tuple with the pack format it has the marker code + 10 bytes, but the raw structure is just 10?

foundationdb/src/transaction.rs
```diff
@@ -464,6 +464,51 @@ impl Transaction {
...
+        TrxVersionstamp {
+            inner: FdbFuture::new(f),
+        }
+    }
+
+    /// The transaction obtains a snapshot read version automatically at the time of the first call
+    /// to fdb_transaction_get_*() (including this one) and (unless causal consistency has been
+    /// deliberately compromised by transaction options) is guaranteed to represent all
+    /// transactions which were reported committed before that call.
+    pub fn get_read_version(&self) -> TrxReadVersion {
```
bluejekyll: I haven’t read the docs enough to understand this yet. Maybe you can explain this to me... what is the difference between this version and the VersionStamp?

I.E. why is this i64 and in the other context a large set of 10 or 12 bytes?

foundationdb/src/transaction.rs
yjh0502: It seems that API versionstamp and Tuple versionstamp has different size: https://forums.foundationdb.org/t/implementing-versionstamps-in-bindings/250

foundationdb/src/transaction.rs
yjh0502: It seems that a versionstamp consists of 8-byte long transaction version and 2-byte transaction batch order. `get_read_version` returns transcation version only, which increments roughly one per each microsecond (1,000,000 hz) [1]

[1] https://forums.foundationdb.org/t/version-length-is-53-bits-enough/260/5

foundationdb/src/transaction.rs
bluejekyll: We’ll need to be careful with this one :)

foundationdb/src/transaction.rs
bluejekyll: That link didn’t explain a ton. So negative number means what? Or is u64 just as valid?

bluejekyll:
I’ll meet he this in. We can revisit later if there’s any additional type safety we want.

PR #37, opened 2018-04-26
(WIP) Tuple

bluejekyll:
Just the initial types defined. Lots of encoding to go...

fixes #35

Issue #39, opened 2018-04-27
Integrate bindingtester


yjh0502:
bindingtester doc:
https://github.com/apple/foundationdb/blob/master/bindings/bindingtester/spec/bindingApiTester.md

bindingtester runner:
https://github.com/apple/foundationdb/blob/master/bindings/bindingtester/bindingtester.py

yjh0502:
bindingtester instruction itself is encoded to tuple, so tuple type should be supported to integrate bindingtester.
https://github.com/apple/foundationdb/blob/master/bindings/go/src/_stacktester/stacktester.go#L437
related issue: https://github.com/bluejekyll/foundationdb-rs/issues/35

yjh0502:
I'm now working on it.

PR #40, opened 2018-04-27
support rust-style tuple

yjh0502:
A goal of this PR is enable rust-style, type-safe tuple decoding, like

```rust
let data: &[u8] = &[2, 104, 101, 108, 108, 111, 0, 1, 119, 111, 114, 108, 100, 0];
let (v1, v2): (String, Vec<u8>) = Tuple::decode(data).expect("Failed to decode tuple");
```

I think at the end there will be two types of tuple API
 - encode/decode arbiraty type of tuple (`Tuple` type in #37). It is similar with `serde_json::Value`, which can be used to decode/encode arbitrary shape of tuples.
 - strong-type tuple. It it similar with `serde_json::from_str`, which decodes the data with expected types, and returns error if type mismatches.

As @bluejekyll is working for on first type of API, I wrote some code for second type of API.

Here's list of works before merging this PR
 - [ ] clean-up names. I borrowed name `Tuple` and `Single` from #37, so it should be changed to something else.
 - [x] floating-point types.
 - [x] integer encoding.
 - [x] support for tuple encoding.
 - [ ] nested tuple.
 - [ ] clean-up constants.

Here's some future works
 - Supporting versionstamp.
 - ...


foundationdb/src/tuple.rs
```diff
@@ -0,0 +1,364 @@
...
+
+					Ok(out)
+				}
+            }
+        )+
+    }
+}
+
+tuple_impls! {
+    1 => (0 T0)
```
bluejekyll: Very cool!

foundationdb/src/tuple.rs
```diff
@@ -0,0 +1,364 @@
+use byteorder::{self, ByteOrder};
+use std;
+use std::io::Write;
+
+#[derive(Debug, Clone, Copy)]
+pub enum TupleError {
```
bluejekyll: Should we make this a Failure?

foundationdb/src/tuple.rs
```diff
@@ -0,0 +1,364 @@
...
+            return Err(TupleError::InvalidData);
+        }
+        Ok(val)
+    }
+}
+
+impl Single for bool {
+    fn encode<W: Write>(&self, mut w: W) -> std::io::Result<()> {
+        if *self {
+            w.write_all(&[0x27])
```
bluejekyll: We should define an enum for these codes, you can copy that from my initial tuple impl...

bluejekyll: See the SingleType from that with it’s associated to_code impl.

bluejekyll:
@yjh0502 I really like the work here. Do you want to land this, and then we can work on this together?

yjh0502:
@bluejekyll I added an patches which is required for #39 .

bluejekyll:
Shall I merge this in as is?

bluejekyll:
> clean-up names. I borrowed name Tuple and Single from #37, so it should be changed to something else.

we can choose whatever here. Probably stick with Tuple, and then for the inner unit, just Item, maybe, to be consistent with Iterator, etc.

PR #43, opened 2018-04-29
add `bindingtester`, #39

yjh0502:
Now our implementation passes `bindingtester`!

Here's bindingtester output.

```
root@c51538127804:/opt/foundationdb# RUST_BACKTRACE=1 ./bindings/bindingtester/bindingtester.py --no-threads --seed 100 ./bindingtester

Creating test at API version 510
Generating scripted test at seed 100 with 100 op(s) and 1 concurrent tester(s)...

Inserting test into database...

Running tester './bindingtester test_spec 510'...

Reading results from '('tester_output', 'results')'...

Comparing results from '('tester_output', 'results')'...

Test with seed 100 and concurrency 1 had 0 incorrect result(s) and 0 error(s) at API version 510
Completed scripted test with random seed 100 and 100 operations
```

 - I added `tuple::SingleValue` and `tuple::TupleValue` to generate bindingtester. Those implementation is somewhat ad-hoc so it could be replaced by one from @bluejekyll 
 - Main loop could be improved to simulate multiple threads on single event loop. I used tons of `wait` to make our bindingtester work. After removing `wait` to make `StackMachine::run` fully asynchronous, we can add support for threading by running multiple `StackMachine` on a event loop without complicating structure with `thread::spawn`.
 - It would be better to integrate bindingtester with CI pipeline. It seems that we need to build foundationdb before running bindingtester, which is quite slow, so I'm not sure it's worth for now.

Here's an instruction to run bindingtester on local environment.

 - build `foundationdb` with https://github.com/apple/foundationdb#compiling-from-source. `master` branch updates API version to 520, so I used `5.1.7` tag to use api version `510`.
 - run `bindingtester.py` with our bindingtester runner.
```sh
RUST_BACKTRACE=1 ./bindings/bindingtester/bindingtester.py --no-threads --seed 100 ./bindingtester
```


yjh0502:
I'm checking invalid memory access on CI
https://travis-ci.org/bluejekyll/foundationdb-rs/jobs/372609487#L734

bluejekyll:
This is great news! I haven’t had the time to put into this you have. Really awesome work! I’m impressed with the speed you’ve gotten us to this point.

I’ll try and spend some time today getting an integration test setup to run this on Travis.

PR #45, opened 2018-04-29
add bindingtester to travis

bluejekyll:
This adds the `scripts/run_bindingtester.sh` which:

1. installs the `mono` dependency needed for the python build
2. builds the `bindingtester` bin
3. syncs the foundationdb repo
4. builds the python bindings
5. runs the `bindingtester.py` script


bluejekyll:
Ok, I think this is good.

bluejekyll:
Example successful log

PR #46, opened 2018-04-30
a type safe SingleType and conversions

bluejekyll:
This changes all the u8 values to the SingleType enum from the other PR. I think this is cleaner.

### DIFF
--- a/foundationdb/src/tuple.rs
+++ b/foundationdb/src/tuple.rs
@@ -13,78 +13,81 @@ use std::io::Write;
 
 use byteorder::{self, ByteOrder};
 
+/// Various tuple types
+const EMPTY: u8 = 0x00;
+const BYTES: u8 = 0x01;
+const STRING: u8 = 0x02;
+const NESTED: u8 = 0x05;
+const INTZERO: u8 = 0x14;
+const POSINTEND: u8 = 0x1d;
+const NEGINTSTART: u8 = 0x0b;
+const FLOAT: u8 = 0x20;
+const DOUBLE: u8 = 0x21;
+const FALSE: u8 = 0x26;
+const TRUE: u8 = 0x27;
+const UUID: u8 = 0x30;
+const VERSIONSTAMP: u8 = 0x33;
+
 #[derive(Debug, Fail)]
 pub enum TupleError {
     #[fail(display = "Unexpected end of file")]
     EOF,
-    #[fail(display = "Invalid type")]
-    InvalidType(u8),
+    #[fail(display = "Invalid type: {}", value)]
+    InvalidType { value: u8 },
     #[fail(display = "Invalid data")]
     InvalidData,
 }
 
 type Result<T> = std::result::Result<T, TupleError>;
 
-/// Various tuple types
-#[derive(Clone, Copy, Eq, PartialEq)]
-enum SingleType {
-    Empty = 0x00,
-    Bytes = 0x01,
-    String = 0x02,
-    Nested = 0x05,
-    IntZero = 0x14,
-    PosIntEnd = 0x1d,
-    NegIntStart = 0x0b,
-    Float = 0x20,
-    Double = 0x21,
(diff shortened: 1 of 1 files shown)


foundationdb/src/tuple.rs
```diff
@@ -8,20 +8,81 @@
...
+    Double = 0x21,
+    False = 0x26,
+    True = 0x27,
+    Uuid = 0x30,
+    VersionStamp = 0x33,
+}
+
+impl SingleType {
+    /// Tries to convert a u8 into a SingleType
+    fn try_from(value: u8) -> Result<Self> {
```
yjh0502: How about implementing `std::convert::TryFrom` trait?
https://doc.rust-lang.org/std/convert/trait.TryFrom.html

foundationdb/src/tuple.rs
```diff
@@ -404,39 +455,32 @@ impl Single for SingleValue {
...
             }
-            val => {
-                if val >= 0x0b && val <= 0x1d {
-                    let (v, offset) = Single::decode(buf)?;
-                    Ok((SingleValue::Int(v), offset))
-                } else {
-                    unimplemented!("unknown code: 0x{:02x}", val);
-                }
-            }
+            _ => unimplemented!(),
```
yjh0502: It will break bindingtester tests, as it requires encoding/decoding tuple integer type.

foundationdb/src/tuple.rs
```diff
@@ -8,20 +8,81 @@
...
+    InvalidType(u8),
+    #[fail(display = "Invalid data")]
     InvalidData,
 }
 
 type Result<T> = std::result::Result<T, TupleError>;
 
+/// Various tuple types
+#[derive(Clone, Copy, Eq, PartialEq)]
+enum SingleType {
```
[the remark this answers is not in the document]
yjh0502: It seems that there's little benefit for using enum for this case. Using enumuration compicates things by enforcing `as u8` cast everywhere, while `try_from` is only used on `Single::decode`. How about using `const TYPE_EMPTY: u8 = 0x00`?

foundationdb/src/tuple.rs
bluejekyll: yes, good catch, I need to go back and fix this.

foundationdb/src/tuple.rs
bluejekyll: TryFrom is nightly only right now, is it not?

yjh0502: I thought it landed on stable, but not yet as yoy told. Sorry for confusion.

foundationdb/src/tuple.rs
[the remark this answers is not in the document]
bluejekyll: > How about using `#[repr(u8)]`?

Fantastic recommendation!

> It seems that there's little benefit for using enum for this case

The benefits I see are this:
- each SingleType code is unique and should not overlap, enum captures that
- in match clauses we get uncovered case errors during compilation.
- It allows us to associate custom functionality, which we'd have to do with a Trait + impl for u8, this is a little cleaner than that.

If you think this generally makes the code harder to follow, and doesn't do much to reduce errors, then I'm happy to switch to named const's, personally I like using this patter in cases like this, but I'm not going to push it.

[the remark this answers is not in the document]
yjh0502: I think we'd better to go with named constant.

The second one is the most beneficial one as it catches tons of human errors. However in this case, we need to use wildcard match anyway because range of header values from `NegIntStart` to `PosIntEnd` maps to integer type.
The first one could be covered by compiler warnings on match clause. For example,
https://play.rust-lang.org/?gist=8769a3fcd215ba9a08b0b3ec32e5654d&version=stable&mode=debug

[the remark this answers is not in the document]
bluejekyll: Ok, I've converted to named consts.

PR #47, opened 2018-05-01
Nested tuple

bluejekyll:
this builds off #46 

I think it's correct :) I need to find some other validations from one of the other binging tests...

### DIFF
--- a/foundationdb/src/tuple.rs
+++ b/foundationdb/src/tuple.rs
@@ -141,7 +141,7 @@ impl Single for () {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct Uuid([u8; 16]);
 
 impl Single for Uuid {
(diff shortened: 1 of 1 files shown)


foundationdb/src/tuple.rs
```diff
@@ -216,6 +216,37 @@ impl Single for String {
...
+    fn decode(buf: &[u8]) -> Result<(Self, usize)> {
+        if buf.len() < 2 {
+            return Err(TupleError::EOF);
+        }
+
+        NESTED.expect(buf[0])?;
+
+        let mut tuples = Vec::new();
+        let mut idx = 1;
+        while buf[idx] != EMPTY {
```
yjh0502: We should check index boundary condition. Following code panics with out-of-bound

```
SingleValue::decode(&[NESTED, INTZERO])
```

yjh0502:
By the way I rebased and force-pushed the branch to fix conflict before review.

bluejekyll:
This should guard against the out-of-bounds issue. I also changed the EMPTY type to NIL...

At some future point, we may want to split decode and encode into different traits so that we can more generically implement encode, where we can't with decode.

bluejekyll:
let's not merge this yet. I want to see if there's another option to better deal with TupleValue...

PR #50, opened 2018-05-02
add `Transaction::add_conflict_range`, #20

yjh0502:


foundationdb/src/transaction.rs
```diff
@@ -561,6 +561,31 @@ impl Transaction {
...
+
+    /// Adds a conflict range to a transaction without performing the associated read or write.
+    ///
+    /// # Note
+    ///
+    /// Most applications will use the serializable isolation that transactions provide by default
+    /// and will not need to manipulate conflict ranges.
+    pub fn add_conflict_range(
+        &self,
+        begin: &[u8],
```
bluejekyll: I had a thought last night about these borrows. Because they are passed in as pointers in the transaction, do they actually have to live until the transaction commit?

I opened #55 for that question.

If i have some time I’ll experiment with that.

bluejekyll:
This looks good. Thanks!

PR #51, opened 2018-05-02
tuple: support nested Empty, fix overflow on int

yjh0502:
Empty value in nested tuple should be escaped with following `0xff` byte, to disambiguate itself from end-of-tuple marker `0x00`.

Here's Golang implementation:

https://github.com/apple/foundationdb/blob/master/bindings/go/src/fdb/tuple/tuple.go#L180-L182
https://github.com/apple/foundationdb/blob/master/bindings/go/src/fdb/tuple/tuple.go#L318-L326


bluejekyll:
Ah. Very nice. I was worried jeering how tuple and ranges would work. I think this answers that?

Issue #36, opened 2018-04-25
Higher-layer API: Subspace


yjh0502:
Reference implementation in Go.
https://github.com/apple/foundationdb/blob/master/bindings/go/src/fdb/subspace/subspace.go

It is an API design issue rather than implementation issue.

yjh0502:
Tuple should be implemented at first: https://github.com/bluejekyll/foundationdb-rs/issues/35

PR #54, opened 2018-05-02
add Subspace, #36

yjh0502:
API and documentation is borrowed from Golang implementation:
https://godoc.org/github.com/apple/foundationdb/bindings/go/src/fdb/subspace

### DIFF
--- a/foundationdb/src/subspace.rs
+++ b/foundationdb/src/subspace.rs
@@ -1,0 +1,158 @@
+// Copyright 2018 foundationdb-rs developers, https://github.com/bluejekyll/foundationdb-rs/graphs/contributors
+// Copyright 2013-2018 Apple, Inc and the FoundationDB project authors.
+//
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+//! subspace provides a convenient way to use FoundationDB tuples to define namespaces for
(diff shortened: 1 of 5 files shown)


bluejekyll:
Overall this looks good.

I have to wonder what our long-term strategy should be with subspace, range keys, and tuple. I can’t help but feel that there’s a better abstraction we can come up with here...

foundationdb/src/tuple.rs
```diff
@@ -431,10 +435,10 @@ impl Single for i64 {
 
 pub trait Tuple: Sized {
     fn encode<W: Write>(&self, _w: &mut W) -> std::io::Result<()>;
-    fn encode_to_vec(&self) -> std::io::Result<Vec<u8>> {
+    fn encode_to_vec(&self) -> Vec<u8> {
         let mut v = Vec::new();
-        self.encode(&mut v)?;
-        Ok(v)
+        self.encode(&mut v).unwrap();
```
bluejekyll: Can we change this to expect with a message like, “tuple encoding should never fail”?

foundationdb/src/subspace.rs
```diff
@@ -0,0 +1,158 @@
...
+        if !self.contains(key) {
+            return Err(TupleError::InvalidData);
+        }
+        let key = &key[self.prefix.len()..];
+        Tuple::decode(&key)
+    }
+
+    /// `contains` returns true if the provided key starts with the prefix of this Subspace,
+    /// indicating that the Subspace logically contains the key.
+    pub fn contains(&self, key: &[u8]) -> bool {
```
bluejekyll: This seems misnamed. Perhaps “covers”? Or “is_prefix_of”?

bluejekyll:
I'll merge this in now, I have a good use-case for it in the tutorial.

PR #56, opened 2018-05-03
cleanup subspace

bluejekyll:
There were a couple of things from the subspace commit that I wanted to cleanup, related to unwraps and naming. Let me know what you think.

### DIFF
--- a/foundationdb/src/subspace.rs
+++ b/foundationdb/src/subspace.rs
@@ -68,16 +68,16 @@ impl Subspace {
     /// removed.  `unpack` will return an error if the key is not in this Subspace or does not
     /// encode a well-formed Tuple.
     pub fn unpack<T: Tuple>(&self, key: &[u8]) -> Result<T, TupleError> {
-        if !self.contains(key) {
+        if !self.covers(key) {
             return Err(TupleError::InvalidData);
         }
         let key = &key[self.prefix.len()..];
         Tuple::decode(&key)
     }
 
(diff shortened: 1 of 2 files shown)


foundationdb/src/subspace.rs
```diff
@@ -68,16 +68,16 @@ impl Subspace {
     /// removed.  `unpack` will return an error if the key is not in this Subspace or does not
     /// encode a well-formed Tuple.
     pub fn unpack<T: Tuple>(&self, key: &[u8]) -> Result<T, TupleError> {
-        if !self.contains(key) {
```
yjh0502: I think name `contains` is more consistent with other APIs.

https://apple.github.io/foundationdb/api-python.html#subspaces
https://apple.github.io/foundationdb/javadoc/com/apple/foundationdb/subspace/Subspace.html#contains-byte:A-

bluejekyll: I was aware. I guess we have a choice, be consistent with the others, or be more accurate. I feel like contains is very misleading and wrong. I’m open to other names, like ‘is_prefix_of’

yjh0502: I like `covers` than `is_prefix_of` (BTW I 
prefers `starts_with` in this case, as it is what we are using in the impl). Thanks for the comment! I'll merge after CI.

rushmorem: I also think that `starts_with` would be the perfect name for this. It's also consistent with the `std` library, making it more intuitive.

PR #57, opened 2018-05-03
rename `covers` to `is_start_of`

rushmorem:
EDIT: Updated as per conversation with @bluejekyll and @yjh0502.

### DIFF
--- a/foundationdb/src/subspace.rs
+++ b/foundationdb/src/subspace.rs
@@ -68,16 +68,16 @@ impl Subspace {
     /// removed.  `unpack` will return an error if the key is not in this Subspace or does not
     /// encode a well-formed Tuple.
     pub fn unpack<T: Tuple>(&self, key: &[u8]) -> Result<T, TupleError> {
-        if !self.covers(key) {
+        if !self.is_start_of(key) {
             return Err(TupleError::InvalidData);
         }
         let key = &key[self.prefix.len()..];
         Tuple::decode(&key)
     }
 
-    /// `covers` returns true if the provided key starts with the prefix of this Subspace,
+    /// `is_start_of` returns true if the provided key starts with the prefix of this Subspace,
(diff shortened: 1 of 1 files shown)


foundationdb/src/subspace.rs
```diff
@@ -68,16 +68,16 @@ impl Subspace {
...
         }
         let key = &key[self.prefix.len()..];
         Tuple::decode(&key)
     }
 
-    /// `covers` returns true if the provided key starts with the prefix of this Subspace,
+    /// `starts_with` returns true if the provided key starts with the prefix of this Subspace,
     /// indicating that the Subspace logically contains the key.
-    pub fn covers(&self, key: &[u8]) -> bool {
+    pub fn starts_with(&self, key: &[u8]) -> bool {
```
bluejekyll: My concern, and the reason I didn't use `starts_with`, is that usually the semantics of `starts_with` is to take a prefix and see if the `self` has that at it's start, in this implementation we're checking if the `key` starts with `self`.

This is why I was hesitant to use this before. We discussed an other option, but @yjh0502 was less enthused with it: `is_prefix_of`. Another option would be `is_start_of`.

Does my concern make sense?

yjh0502: I didn't realized that point. I'm okay with all those names, but if I have to choose one of those, I prefer `covers` :)

rushmorem: > in this implementation we're checking if the key starts with self

Ah, I see. That, indeed, makes sense. In that case I think I like `is_prefix_of` or `is_start_of` more than `covers` because to me `covers` is a bit vague. Those 2 explain the relationship more clearly.

bluejekyll: so splitting the difference, should we just land on `is_start_of`?

Issue #58, opened 2018-05-03
segmentation fault on `watch` testcase


yjh0502:
Here's a log

https://gist.github.com/yjh0502/9e4f9f7376edf1e3c3e2c657c996a2bb

rustc version: `rustc 1.25.0 (84203cac6 2018-03-25)`
foundationdb: `5.1.5-1` (matches with CI version)

yjh0502:
Here's a valgrind log with debug symbol. I built a fdbclient with debug symbol.
foudationdb version: 5.1.7 (9ad8d02386d4a6a5efecf898df80f2747695c627)

https://gist.github.com/yjh0502/ae0efed18c056f9a526e924bc34bea5b

PR #59, opened 2018-05-03
fix segfault on watch test, #58

yjh0502:
The segfault (#58) happens when watch outlives database/cluster. It happens befause `TrxWatch` does not keep any refcount to database/cluster. I fixed the issue by making all `FdbFuture` live with either `Transaction` or `Database`.

I found another possible problem while fixing the issue, which is caused by drop order of Rust. Rust drops struct field from top to bottom[1]. If we define `Cluster` prior to `Database`, cluster will be dropped before `Database`, which might cause a problem. `valgrind` does not complain for this case but I changed the field order to just make sure.

[1] https://github.com/rust-lang/rfcs/blob/master/text/1857-stabilize-drop-order.md#tuples-structs-and-enum-variants

### DIFF
--- a/foundationdb/src/transaction.rs
+++ b/foundationdb/src/transaction.rs
@@ -33,8 +33,10 @@ use tuple;
 /// Transactions are also causally consistent: once a transaction has been successfully committed, all subsequently created transactions will see the modifications made by it.
 #[derive(Clone)]
 pub struct Transaction {
-    database: Database,
+    // Order of fields should not be changed, because Rust drops field top-to-bottom, and
+    // transaction should be dropped before cluster.
     inner: Arc<TransactionInner>,
+    database: Database,
 }
 
 /// Converts Rust `bool` into `fdb::fdb_bool_t`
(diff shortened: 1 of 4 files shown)


bluejekyll:
> Rust drops struct field from top to bottom[1].

I don't believe that linked RFC can be relied on, and it doesn't look like there is a path forward on this: https://github.com/rust-lang/rfcs/pull/1857

foundationdb/src/transaction.rs
```diff
@@ -979,16 +1014,19 @@ impl Future for TrxErrFuture {
 
 /// Futures for transaction, which supports retry/backoff with `Database::transact`.
 struct TrxFuture {
-    trx: Option<Transaction>,
+    // Order of fields should not be changed, because Rust drops field top-to-bottom, and future
+    // should be dropped before transaction.
```
bluejekyll: I think we'll need a custom Drop impl on this type. And then you can enforce the order of the dropped components.

PR #60, opened 2018-05-03
refactor the `tuple` module

rushmorem:
Refactors the `tuple` module to:-

* Split `Single` into separate `Encode` and `Decode` traits
* Split `Tuple` into separate `Encode` and `Decode` traits
* Extract "single" types into `tuple::single`
* Rename `TupleError`, `TupleValue`, `SingleValue` and `SingleType` to conform to Rust API guidelines


bluejekyll:
Thanks for the PR!

Issue #61, opened 2018-05-03
Can we use `uuid::Uuid`?


rushmorem:
Is there any particular reason we have a custom `Uuid` type instead of just using the one from the `uuid` crate?

yjh0502:
I think we could add uuid as a optional dependency, as postgres does.

https://github.com/sfackler/rust-postgres/blob/master/postgres-shared/Cargo.toml

rushmorem:
Making it optional makes it possible to support alternative implementations in future in a backwards compatible way, if such a need ever arises. Also some people prefer libraries with as few dependencies as possible. Making external dependencies like that optional, makes it possible for them to turn them off when they don't need them. Considering that it's trivial to implement it as optional, I don't really see any reason not to. We can activate it by default.

PR #62, opened 2018-05-03
use `Uuid` from the `uuid` crate

rushmorem:
Closes https://github.com/bluejekyll/foundationdb-rs/issues/61

### DIFF
--- a/foundationdb/src/tuple/item.rs
+++ b/foundationdb/src/tuple/item.rs
@@ -1,4 +1,6 @@
 use std::{self, io::Write};
+#[cfg(feature = "uuid")]
+use uuid::Uuid;
 
 use tuple::{self, Error, Result};
 use byteorder::{self, ByteOrder};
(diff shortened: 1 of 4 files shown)


foundationdb/src/tuple/single.rs
```diff
@@ -224,7 +228,7 @@ impl Decode for Uuid {
         let mut uuid = [0u8; 16];
         uuid.copy_from_slice(&buf[1..17]);
 
-        Ok((Uuid(uuid), 17))
+        Ok((Uuid::from_random_bytes(uuid), 17))
```
yjh0502: We should use `from_bytes` because `from_random_bytes` forces random uuid variant.

https://docs.rs/uuid/0.6.3/src/uuid/lib.rs.html#743-748

rushmorem: Silly me :) Let me update. Thanks!

foundationdb/src/tuple/item.rs
```diff
@@ -224,7 +228,7 @@ impl Decode for Uuid {
         let mut uuid = [0u8; 16];
         uuid.copy_from_slice(&buf[1..17]);
 
-        Ok((Uuid(uuid), 17))
+        Ok((Uuid::from_uuid_bytes(uuid), 17))
```
rushmorem: @yjh0502 I ended up using `from_uuid_bytes` because `from_bytes` doesn't check the first byte like you do.

PR #63, opened 2018-05-03
ignore Cargo.lock

rushmorem:
It has no business in a library :)


bluejekyll:
Thanks. I suppose this is unnecessary. I always have disagreed with the Rust FAQ on this, as I think it's useful to know what versions have been tested when issues are encountered by others. We can remove it, but I don't think it has much of a negative, other than causing annoying conflicts on merge every so often.

rushmorem:
I beg to differ.

> I think it's useful to know what versions have been tested when issues are encountered by others.

Yes you are right that knowing the exact versions being used by a library can indeed help when debugging. However, the fact that the Rust community uses SemVer and that libraries specify the range of versions they support in `Cargo.toml` does mitigate the risk somewhat. Say I notice an issue in my program when I'm depending on certain libraries. The first thing I might do is `cargo update` to rebuild with latest versions of those libraries to see if the issue would persist. After all, it's a waste of everybody's time to try and debug a solved issue, right? If the issue persists at that point then it's not a big deal because I know what the latest versions at that point are, right?

> I don't think it has much of a negative, other than causing annoying conflicts on merge every so often.

`git` makes dealing with those merge conflicts trivial. Especially for a file like `Cargo.lock`. I think pinning down library dependencies to specific versions can cause unnecessary dependency hell. That's where the issue is.

Take, for example, a library like `ring`. It currently doesn't allow programs to depend on multiple versions of it. The latest version currently is `0.12.1`. Say a program has more than one dependency that also depends on `ring`. If all those libraries support `ring = "0.12"` or something like that, that's not a problem, because only `0.12.1` will be pulled down and built even if the developers of those libraries had different versions of `ring` in their lock files. However, if at least one of those dependencies has a `Cargo.lock` that locks `ring` down to `0.12.0` then the program won't build.

This problem doesn't only apply to libraries that disallow multiple versions. Traits defined in external libraries sometimes appear not to be implemented for certain types when they are, due to mismatched versions.

bluejekyll:
Please correct me if I’m wrong here, but it’s my understanding that Cargo.lock is ignored via downstream library dependencies, and only the Cargo.toml restrictions are evaluated.

rushmorem:
> Please correct me if I’m wrong here, but it’s my understanding that Cargo.lock is ignored via downstream library dependencies, and only the Cargo.toml restrictions are evaluated.

I'm not sure. If that's the case, then leaving `Cargo.lock` in the repo is, indeed, harmless. I always assumed that if it's present, it would be used instead even for downstream dependencies.

bluejekyll:
> If we exclude it in Cargo.toml but leave it checked in for CI it means we might even ship broken code without knowing it.

That's a valid point... ok, removing.

PR #64, opened 2018-05-03
rename `single` to `item`

rushmorem:
I think `item` makes it more clear that these are tuple items.
Before I actually read the code, it wasn't clear to me what
`single` was. I thought maybe it was a tuple with a single item.

I think renaming it this way makes the API a bit easier to
understand. Thoughts? @bluejekyll @yjh0502


bluejekyll:
yes, this is good.

Issue #31, opened 2018-04-25
Implement FoundationDB tutorial as example


bluejekyll:
https://apple.github.io/foundationdb/class-scheduling-java.html

yjh0502:
It seems that tuple layer should be implemented first: https://github.com/bluejekyll/foundationdb-rs/issues/35

bluejekyll:
Yeah, I started working on a Tuple as part of this... I will break that out for a separate review. Are you already doing that?

PR #65, opened 2018-05-04
Class scheduling tutorial

bluejekyll:
Fixes: #31 

I need to go through and validate this still. It was derived from the Java tutorial, so it definitely can be more Rusty.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,5 +57,6 @@ script:
   - cargo test -p foundationdb-gen --all-features
   - cargo test -p foundationdb --all-features
   - cargo test -p foundationdb-bench --all-features
+  - cargo run -p foundationdb --example class-scheduling --all-features 
 
 # after_success: cargo kcov here...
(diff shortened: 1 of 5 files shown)


bluejekyll:
Ok, before pushing this, I want to cleanup some of the interfaces... not too much though, as I want it to basically mimic the Tutorial from FoundationDB.

bluejekyll:
A few more changes. I need to cleanup the `FdbError` for use in the `Database::transact`, almost finished that refactor, but am out of time tonight.

Once that's done, I can convert to using `transact` as required in the tutorial.

bluejekyll:
After #77 lands, I think this is ready. And then I say we publish the next release.

bluejekyll:
This depends on #77

bluejekyll:
Ok, I might go back and interleave the tutorial with the code later.

rushmorem:
OK

PR #70, opened 2018-05-04
add benchmark

yjh0502:
Add a benchmark to check if 1) our library works well under load, and 2) measure an overhead on top of C API, and find a bottleneck if exists.

On my local enviroment (i5-4690 with Samsung SSD 850 PRO 256GB), the library can saturate single foundationdb server with single event loop. I found some bottleneck while benchmarking and fixed some of them.

A benchmark result shows that our implementation has ~10% overhead on top of raw C API, including time spent by benchmark program itself. Here's final `perf` output.

```
# Total Lost Samples: 17
#
# Samples: 1K of event 'cycles:ppp'
# Event count (approx.): 35121043738
#
# Overhead  Shared Object
# ........  ..................
#
    88.85%  libfdb_c.so
     9.12%  fdb-bench
     0.73%  [kernel.kallsyms]
     0.65%  libc-2.27.so
     0.44%  ld-2.27.so
     0.10%  [vdso]
     0.05%  [nf_conntrack]
     0.05%  [unknown]
     0.01%  libpthread-2.27.so
     0.01%  [ip_tables]
```

I introduced `foundationdb-bench` crate because of handling dependencies. It might be better to move `fdb-bench` to `foundationdb/benches`, `examples`, or `tests`. In that case we should move dependencies to `[dev-dependencies]` section on `foundationdb` crate. I'm not sure which one is better, and I don't have any strong opinion on that.


bluejekyll:
I wonder how much of that overhead is related to key construction, etc. I think we can improve quite a bit there now that the encoders and decoders are split.

Cargo.toml
```diff
@@ -1,4 +1,8 @@
 [workspace]
 members = ["foundationdb", 
            "foundationdb-sys",
-           "foundationdb-gen"]
+           "foundationdb-gen",
+           "foundationdb-bench"]
```
bluejekyll: I think this is fine, unless you see cross-over that would be useful.

foundationdb-bench/src/bin/fdb-bench.rs
```diff
@@ -0,0 +1,181 @@
...
+    key_len: usize,
+    #[structopt(long = "val-len", default_value = "100")]
+    val_len: usize,
+}
+
+fn main() {
+    env_logger::init();
+    let opt = Opt::from_args();
+
+    let network = fdb_api::FdbApiBuilder::default()
```
bluejekyll: you can just use `foundationdb::init()` for the defaults...

bluejekyll:
I think we should at least build this as part of the travis script so that it doesn't fall behind.

Issue #68, opened 2018-05-04
Rename `item::Value::Str` variant to `item::Value::String`


rushmorem:
By convention, in Rust, if an enum variant stores a `String`, it is named `String`. `Str` is usually used for `&str`. See [Serde](https://docs.rs/serde_json/1.0.17/serde_json/enum.Value.html) for example.

rushmorem:
I noticed that after https://github.com/bluejekyll/foundationdb-rs/pull/67, our API for the tuple layer is shaping up quite nicely. I think this is a good opportunity to give it one more (and hopefully last) push. The item layer is only left with `item::Value` publicly available. While we are at it, I propose we do the following to cleanup:-

* Make `tuple::item` private
* Reexport `item::Value` as `tuple::Item`
* Rename the other variants of `item::Value` that use `C` terminology like `Double` to the names of the types that Rust uses, like `F64`. I think this will make the API easier to grasp for users who are not familiar with `C`. In the process, we will rename `Int(i64)` to `I64(i64)`. This will make it easier to introduce more types in future, eg `I128(i128)`, in a backwards compatible way. Which brings me to my last item.
* Make `item::Value` non-exhaustive to make sure we can add more types in a backwards compatible way in future

PR #71, opened 2018-05-04
Clean up the `tuple` module

rushmorem:
As proposed in https://github.com/bluejekyll/foundationdb-rs/issues/68.


rushmorem:
Yeah, I like that. I have updated the PR.

rushmorem:
I've also implemented `Encode` for `&str` in this PR because that was going to require yet another breaking change.

bluejekyll:
Awesome! Thanks!

rushmorem:
Ah, you are absolutely right! I didn't think this through. I didn't need to put it in the variant. I will fix this shortly.

Issue #72, opened 2018-05-05
impl `Sync`/`Send` for `Cluster`/`Database`?


yjh0502:
It seems that C API is thread-safe unless explicitly mentioned.
https://github.com/apple/foundationdb/blob/master/fdbclient/ThreadSafeTransaction.actor.cpp#L30-L31

Rust raw pointers are not `Sync`/`Send` by default so `Database` cannot be sent to other thread although `DatabaseInner` is wrapped by `std::sync::Arc`. We can fix the problem by marking `DatabaseInner` and friends as `Sync`/`Send` explicitly.

bluejekyll:
This is done, yest?

yjh0502:
Yes

PR #73, opened 2018-05-05
bench: add support for threading/batching

yjh0502:
Add support for threading/batching.
 - threading: use multiple threads to inserts key/value pairs, with `futures-cpupool`.
 - batching: write multiple key/value pairs on single transaction.

I also made RNG deterministic, so that benchmark will rewrite existing keys with same parameters. It makes benchmark result more stable.

Main issue on this PR is `Send`/`Sync`. As I mentioned on the issue #72 , it seems that cluster/database/transaction could be shared between multiple threads, so I marked those as a `Send`/`Sync`.

### DIFF
--- a/foundationdb/src/future.rs
+++ b/foundationdb/src/future.rs
@@ -48,8 +48,6 @@ impl Drop for FdbFuture {
     }
 }
 
-unsafe impl Send for FdbFuture {}
-
 impl futures::Future for FdbFuture {
     type Item = FdbFutureResult;
     type Error = FdbError;
--- a/foundationdb/src/future.rs
+++ b/foundationdb/src/future.rs
@@ -48,6 +48,8 @@ impl Drop for FdbFuture {
     }
 }
 
+unsafe impl Send for FdbFuture {}
+
 impl futures::Future for FdbFuture {
     type Item = FdbFutureResult;
     type Error = FdbError;
--- a/foundationdb/src/transaction.rs
+++ b/foundationdb/src/transaction.rs
@@ -613,6 +613,8 @@ impl Drop for TransactionInner {
         }
     }
 }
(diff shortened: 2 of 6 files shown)


foundationdb/src/transaction.rs
```diff
@@ -613,6 +613,8 @@ impl Drop for TransactionInner {
         }
     }
 }
+unsafe impl Send for TransactionInner {}
```
bluejekyll: I can believe that Database is threadsafe, but is transaction as well?

foundationdb/src/future.rs
```diff
@@ -48,6 +48,8 @@ impl Drop for FdbFuture {
     }
 }
 
+unsafe impl Send for FdbFuture {}
```
bluejekyll: I worry about marking FdbFuture Send. I can imagine a future where we re-implement the the C++ client code in native Rust and directly on Tokio, etc. In that context, Future may not be send, right?

That is, my concern is that we may paint ourselves into a corner by making FdbFuture Send now. So I might suggest holding off on making this Send (unless there is a need to do this now).

foundationdb/src/transaction.rs
yjh0502: There's an implication in a documentation.

https://docs.rs/foundationdb/0.1.0/foundationdb/options/enum.TransactionOption.html

NextWriteNoWriteConflictRange
Care needs to be taken when using this option on a transaction that is shared between multiple threads. When setting this option, write conflict ranges will be disabled on the next write operation, regardless of what thread it is on.

foundationdb/src/future.rs
yjh0502: I agree. I'll add a patch to fix it.

PR #74, opened 2018-05-05
encode some borrowed types

rushmorem:
* Reverts previous `&str` encoding. See https://github.com/bluejekyll/foundationdb-rs/pull/71#issuecomment-386812699.
* Implements `Encode` for `&str` and `&[u8]`
* Adds a blanket `Encode` implementation for all borrowed types that implement `Encode`


rushmorem:
No problem :)

Issue #49, opened 2018-05-01
tuple: Support Rust-style nested tuple


yjh0502:
https://github.com/bluejekyll/foundationdb-rs/pull/47#issuecomment-385746621

bluejekyll:
I think there is nothing to do here... There is one thing that would be cool, but I'm not sure what we can do here, right now you can't decode into something that's comparible with borrowed tuples, example:

`("str", (4,2)).encode_to_vec() == ("str".to_string(), (4,2)).encode_to_vec()`

but this is not true:

`("str", (4,2)) != <(String, (i64, i64))>.decode_full(&[...])`

Not sure what we can do there...

rushmorem:
> Right now you can't decode into something that's comparable with borrowed tuples

This is a limitation from how Rust compares tuples.

```rust
("str", (4,2)) != ("str".to_string(), (4,2))
```

bluejekyll:
I think we should drop the single tuple variant generate in the tuple module: `(0 T0)`

It would remove this confusion: as far as I can tell `(1,).encode() == 1.encode()`, we really need to validate that. Especially b/c I'm fairly certain that `(1, (1,)).encode() != (1, 1).encode()` and that's a little awkward...

PR #77, opened 2018-05-06
Allow the association of other errors to FdbError

bluejekyll:
To make `transact` a little more flexible, I've added ErrorKind to FdbError.
- renamed FdbError -> Error
- introduced ErrorKind
- introduced a `From<failure::Error>` for ease of conversion
- transact now requires `Into<foundationdb::error::Error>` for it's future error result
- Adds deref `Vec<Element>` to tuple

### DIFF
--- a/foundationdb-gen/src/bin/foundationdb-options-gen.rs
+++ b/foundationdb-gen/src/bin/foundationdb-options-gen.rs
@@ -111,7 +111,7 @@ impl FdbScope {
 
         let mut s = String::new();
         s += &format!(
-            "pub unsafe fn apply(&self{}) -> std::result::Result<(), error::FdbError> {{\n",
+            "pub unsafe fn apply(&self{}) -> std::result::Result<(), error::Error> {{\n",
             first_arg
         );
         s += "let code = self.code();\n";
@@ -159,7 +159,7 @@ impl FdbScope {
         }
 
         s += "};\n";
-        s += "if err != 0 { Err(error::FdbError::from(err)) } else { Ok(()) }\n";
+        s += "if err != 0 { Err(error::Error::from(err)) } else { Ok(()) }\n";
         s += "}\n";
 
         s
--- a/foundationdb/src/bin/bindingtester.rs
+++ b/foundationdb/src/bin/bindingtester.rs
@@ -6,7 +6,7 @@ extern crate log;
 
 use std::collections::HashMap;
 
-use fdb::error::FdbError;
+use fdb::error::Error;
 use fdb::keyselector::KeySelector;
 use fdb::tuple::*;
 use fdb::*;
@@ -251,7 +251,7 @@ impl Instr {
     }
 }
 
-type StackFuture = Box<Future<Item = (Transaction, Vec<u8>), Error = FdbError>>;
+type StackFuture = Box<Future<Item = (Transaction, Vec<u8>), Error = Error>>;
 struct StackItem {
     number: usize,
     // TODO: enum
@@ -350,7 +350,7 @@ impl StackMachine {
         }
     }
 
-    fn fetch_instr(&self) -> Box<Future<Item = Vec<Instr>, Error = FdbError>> {
+    fn fetch_instr(&self) -> Box<Future<Item = Vec<Instr>, Error = Error>> {
         let db = self.db.clone();
 
         let prefix = (self.prefix.clone(),);
@@ -366,7 +366,7 @@ impl StackMachine {
                         let instr = Instr::from(kv.value());
                         instrs.push(instr);
                     }
-                    Ok(instrs)
+                    Ok::<_, Error>(instrs)
                 });
             f
         });
@@ -420,7 +420,7 @@ impl StackMachine {
 
     fn push_fut<F>(&mut self, number: usize, fut: F)
     where
-        F: Future<Item = (Transaction, Vec<u8>), Error = FdbError> + 'static,
+        F: Future<Item = (Transaction, Vec<u8>), Error = Error> + 'static,
     {
         let item = StackItem {
             number,
@@ -494,7 +494,7 @@ impl StackMachine {
             OnError => {
                 let code: i64 = self.pop_item();
                 let trx0 = trx.clone();
-                let f = trx.on_error(FdbError::from(code as i32))
+                let f = trx.on_error(Error::from(code as i32))
                     .map(move |_| (trx0, b"RESULT_NOT_PRESENT".to_vec()));
                 self.push_fut(number, f);
             }
@@ -603,7 +603,7 @@ impl StackMachine {
                             key.to_vec().encode(&mut out).expect("failed to encode");
                             value.to_vec().encode(&mut out).expect("failed to encode");
                         }
-                        Ok(out)
+                        Ok::<_, Error>(out)
                     })
                     .map(|out| (trx0, out));
 
--- a/foundationdb/src/bin/bindingtester.rs
+++ b/foundationdb/src/bin/bindingtester.rs
@@ -177,7 +177,7 @@ impl Instr {
         use InstrCode::*;
 
         let tup: Tuple = Decode::decode_full(data).unwrap();
-        let cmd = match tup.0[0] {
+        let cmd = match tup[0] {
             Element::String(ref s) => s.clone(),
             _ => panic!("unexpected instr"),
         };
@@ -191,7 +191,7 @@ impl Instr {
 
         let code = match cmd {
             "PUSH" => {
-                let data = tup.0[1].encode_to_vec();
+                let data = tup[1].encode_to_vec();
                 Push(data)
             }
             "DUP" => Dup,
--- a/foundationdb/src/database.rs
+++ b/foundationdb/src/database.rs
@@ -65,39 +65,55 @@ impl Database {
     /// It might retry indefinitely if the transaction is highly contentious. It is recommended to
     /// set `TransactionOption::RetryLimit` or `TransactionOption::SetTimeout` on the transaction
     /// if the task need to be guaranteed to finish.
-    pub fn transact<F, Fut, Item, E>(&self, f: F) -> Box<Future<Item = Fut::Item, Error = Error>>
+    pub fn transact<F, Fut, Item, E>(&self, f: F) -> Box<Future<Item = Fut::Item, Error = E>>
     where
         F: FnMut(Transaction) -> Fut + 'static,
         Fut: Future<Item = Item, Error = E> + 'static,
         Item: 'static,
-        E: Into<Error>,
+        E: From<Error> + Send + Sync + 'static,
     {
+        // This will be used to temporarily capture the internal error of the FnMut, F.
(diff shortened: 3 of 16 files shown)


foundationdb/src/bin/bindingtester.rs
```diff
@@ -603,7 +603,7 @@ impl StackMachine {
                             key.to_vec().encode(&mut out).expect("failed to encode");
                             value.to_vec().encode(&mut out).expect("failed to encode");
                         }
-                        Ok(out)
```
bluejekyll: This is unfortunate, not sure if there is a way to get rid of this required type ascription...

foundationdb/src/error.rs
```diff
@@ -8,91 +8,160 @@
...
+        error_code: i32,
+        /// The error string as defined by FoundationDB
+        error_str: &'static str,
+    },
+    /// Other errors not directly attributed to the FoundationDB layers
+    #[fail(display = "FoundationDB-RS error: {}", error)]
+    Other {
+        /// The cause of this error
+        error: failure::Error,
+    },
```
bluejekyll: at some point we should probably add TupleError here...

foundationdb/src/options.rs
```diff
@@ -147,7 +147,7 @@ impl NetworkOption {
             }
         }
     }
-    pub unsafe fn apply(&self) -> std::result::Result<(), error::FdbError> {
+    pub unsafe fn apply(&self) -> std::result::Result<(), error::Error> {
```
bluejekyll: I think we have a gap in our testing here... this should be regenerated and the result tested, is that happening now?

foundationdb-gen/src/bin/foundationdb-options-gen.rs
```diff
@@ -111,7 +111,7 @@ impl FdbScope {
 
         let mut s = String::new();
         s += &format!(
-            "pub unsafe fn apply(&self{}) -> std::result::Result<(), error::FdbError> {{\n",
+            "pub unsafe fn apply(&self{}) -> std::result::Result<(), error::Error> {{\n",
```
bluejekyll: while this crate is tested, is there and integration test for the resultant output of this generator? might be a gap in our testing.

foundationdb/src/error.rs
bluejekyll: Done.

foundationdb/src/bin/bindingtester.rs
```diff
@@ -603,7 +603,7 @@ impl StackMachine {
...
+                        Ok::<_, Error>(out)
```
bluejekyll: This is unfortunate, not sure if there is a way to get rid of this required type ascription...

yjh0502:
How about using `Error: From<FdbError>` for `transact`?

The patch will definitely increases usability of the crate, especially when user tries to mix tuple/subspace layers with other APIs. However where will be some inconvenience when user want to write a layer on top of core API. For example, now caller of `Transaction::get` might think they should handle non-core error codes, while they don't need to.

foundationdb/src/database.rs
```diff
@@ -64,18 +65,20 @@ impl Database {
...
     /// set `TransactionOption::RetryLimit` or `TransactionOption::SetTimeout` on the transaction
     /// if the task need to be guaranteed to finish.
-    pub fn transact<F, Fut, Item>(&self, f: F) -> Box<Future<Item = Fut::Item, Error = FdbError>>
+    pub fn transact<F, Fut, Item, E>(&self, f: F) -> Box<Future<Item = Fut::Item, Error = Error>>
     where
         F: FnMut(Transaction) -> Fut + 'static,
-        Fut: Future<Item = Item, Error = FdbError> + 'static,
+        Fut: Future<Item = Item, Error = E> + 'static,
         Item: 'static,
+        E: Into<Error>,
```
yjh0502: How about using `E: From<FdbError>` and use `E::From(fdb_err)` for internal error case? If caller is not emitting error on `F` they can use it with `FdbError`, and if caller uses custom error type with either `failure` or `error-chain`, they can add `From<FdbError>` for there own error type without much effort.

foundationdb/src/tuple/mod.rs
```diff
@@ -196,4 +238,39 @@ mod tests {
 
         assert_eq!("string".encode_to_vec(), ("string",).encode_to_vec());
     }
+
+    #[test]
+    fn test_recursive_tuple() {
```
bluejekyll: I think this test shows that #49 is already ok, @yjh0502

yjh0502: There's an error on the testcase. Here's an output from python fdb. Please ignore a header value of strings, because python2 string is encoded into binary in fdb bindings.
```
>>> [ord(x) for x in fdb.tuple.pack( ("one", ("two", 42)) )]
[1, 111, 110, 101, 0, 5, 1, 116, 119, 111, 0, 21, 42, 0]
```

There's one more edge case on nested tuple: we should handle nested Empty value differently with top-level empty value.
```
>>> [ord(x) for x in fdb.tuple.pack( (None, (None, None)) )]
[0, 5, 0, 255, 0, 255, 0]
```

foundationdb/src/tuple/mod.rs
```diff
@@ -196,4 +238,39 @@ mod tests {
...
+        assert_eq!(
+            &("one", ("two", 42)).encode_to_vec(),
+            &[2, 111, 110, 101, 0, 2, 116, 119, 111, 0, 21, 42]
+        );
+        assert_eq!(
+            &("one", ("two", 42, ("three", 33))).encode_to_vec(),
```
yjh0502:
```
>>> [ord(x) for x in fdb.tuple.pack( ("one", ("two", 42, ("three", 33) ) ) )]
[1, 111, 110, 101, 0, 5, 1, 116, 119, 111, 0, 21, 42, 5, 1, 116, 104, 114, 101, 101, 0, 21, 33, 0, 0]
```

bluejekyll: Ah, you're right... I should have looked more closely. I'll drop this impl from the change.

foundationdb/src/database.rs
```diff
@@ -64,37 +65,55 @@ impl Database {
...
+    pub fn transact<F, Fut, Item, E>(&self, f: F) -> Box<Future<Item = Fut::Item, Error = E>>
     where
         F: FnMut(Transaction) -> Fut + 'static,
-        Fut: Future<Item = Item, Error = FdbError> + 'static,
+        Fut: Future<Item = Item, Error = E> + 'static,
         Item: 'static,
+        E: From<Error> + Send + Sync + 'static,
     {
+        // This will be used to temporarily capture the internal error of the FnMut, F.
+        enum EitherError<E> {
```
bluejekyll: @yjh0502 I think this is needed to temporarily capture the raw error during execution.

foundationdb/src/database.rs
```diff
@@ -64,37 +65,55 @@ impl Database {
...
     /// set `TransactionOption::RetryLimit` or `TransactionOption::SetTimeout` on the transaction
     /// if the task need to be guaranteed to finish.
-    pub fn transact<F, Fut, Item>(&self, f: F) -> Box<Future<Item = Fut::Item, Error = FdbError>>
+    pub fn transact<F, Fut, Item, E>(&self, f: F) -> Box<Future<Item = Fut::Item, Error = E>>
     where
         F: FnMut(Transaction) -> Fut + 'static,
-        Fut: Future<Item = Item, Error = FdbError> + 'static,
+        Fut: Future<Item = Item, Error = E> + 'static,
         Item: 'static,
+        E: From<Error> + Send + Sync + 'static,
```
bluejekyll: @yjh0502 I think this matches what you wanted.

foundationdb/src/database.rs
yjh0502: I think we can do it without introducing another error type.

https://github.com/yjh0502/foundationdb-rs/commit/838ecb2387b7eb57ff890489d100daaf38d7925d

foundationdb/src/database.rs
yjh0502: Yes, this is what I wanted and I didn't catch it before. Sorry for noise.

foundationdb/src/database.rs
bluejekyll: Ah yes... nice work.

foundationdb/src/error.rs
```diff
@@ -8,91 +8,167 @@
...
+    }
+}
+
+impl From<failure::Error> for Error {
+    fn from(error: failure::Error) -> Self {
+        Error::from_other(error)
+    }
+}
+
+impl Error {
```
yjh0502: I'm not sure of this. As I said on previous comment, I believe user should be able to get raw `FdbError` when it is an only error type that user can get, especially for layer developers, which will want to access a API without noise. I agree that that will make user annoying when trying to use, for example `Database::transact` with `Tuple::decode`, but it could be fixed by using `failure` or `error-chain`. I tried to demonstrate an example on here:

https://github.com/yjh0502/foundationdb-rs/commit/838ecb2387b7eb57ff890489d100daaf38d7925d#diff-8fe8d202ec6a71cdcaa54f69083da070R27

yjh0502:
Sorry for late response. I wrote a comment to describe my concern. https://github.com/bluejekyll/foundationdb-rs/pull/77#discussion_r186436286

Except for `Error` type discussion, I'm great with the changes. Could you share your opinions about `FdbError` to `Error` change? If introducing `Error` is mainly for allowing error type other than `FdbError` in `Database::transact`, I think we can do that without `Error`. If the change is mainly for providing single error type over the crate for easy of use and If you think that's a right way to go, I'll follow your decision.

I'm sorry for not expressing my point clear earlier. I should have made my point clear at the first comment.

foundationdb/src/error.rs
bluejekyll: Ok, I've pulled in your changes. That was definitely more more elegant solution, I was one step away, thanks for showing how to do it.

> I'm not sure of this. As I said on previous comment, I believe user should be able to get raw FdbError when it is an only error type that user can get, especially for layer developers, which will want to access a API without noise.

We can drop the Other from this error, but I think having a common error type in the library is generally something people have grown to expect. Granted, after you're changes, and the suggestion to move to `From` on the transact interface the initial reason for this change can be dropped.

But, I think having a common type that covers both the FoundationDB errors and TupleError will be nice for us in the library. So I can drop the `Other`. I'm willing to do either at this point, but I do think this makes the library simpler to work with. It doesn't sound like either of us are strongly in favor of one direction or the other here. If it's helpful, I'll drop it from this PR, but I'm of the opinion that it does make things easier for others...

yjh0502: I got your point, and I agree with your opinion. Thanks!

yjh0502:
@bluejekyll Yes, It's better for me, thanks! Thanks for taking time to share the opinion.

bluejekyll:
Hit the wrong button there. Once tests pass this can be merged.

PR #78, opened 2018-05-07
cleanup `subspace`

rushmorem:
* `Subspace` APIs currently make borrowing mandatory. So instead of simply writing `Subspace::from("start")`, you are forced to write `Subspace::from(&"start")`
* The entire `subspace` module doesn't need to be exported publicly since it's only exposing one struct
* The `From` implementation makes `Subspace::new` redundant

### DIFF
--- a/foundationdb/src/subspace.rs
+++ b/foundationdb/src/subspace.rs
@@ -24,6 +24,14 @@ pub struct Subspace {
     prefix: Vec<u8>,
 }
 
+impl<E: Encode> From<E> for Subspace {
+    fn from(e: E) -> Self {
+        Self {
+            prefix: e.encode_to_vec(),
+        }
+    }
+}
+
 impl Subspace {
     /// `all` returns the Subspace corresponding to all keys in a FoundationDB database.
     pub fn all() -> Subspace {
@@ -34,13 +42,6 @@ impl Subspace {
     pub fn from_bytes(bytes: &[u8]) -> Self {
         Self {
             prefix: bytes.to_vec(),
-        }
-    }
-
-    /// Returns a new Subspace from the provided tuple encodable.
-    pub fn new<T: Encode>(t: T) -> Self {
-        Self {
-            prefix: t.encode_to_vec(),
         }
     }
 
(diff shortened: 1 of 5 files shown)


foundationdb/src/subspace.rs
```diff
@@ -103,21 +106,21 @@ mod tests {
 
     #[test]
     fn sub() {
-        let ss0 = Subspace::new(&(1,));
-        let ss1 = ss0.subspace(&(2,));
+        let ss0: Subspace = (1,).into();
```
yjh0502: It's not a scope of this PR, but is it okay to convert non-tuple element to single-element tuple implicitly? For example now following example results on same subspace
```
let ss1: Subspace = (1,).into();
let ss2: Subspace = 1.into();
assert_eq!(ss1.bytes(), ss1.bytes());
```

If it makes sense, we can make this line shorter with `1.into()`

rushmorem: That's already possible. As long as the element implements `Encode` it can converted. Let me clean up the tests...

yjh0502:
Sorry for confusion. What I want to say is,

It's not a scope of this PR, but I have a question. Is it okay to allow converting a element to single-element tuple implicitly? In the other word, Is it okay to accept a element as a unary tuple for all APIs that accepts tuple type as a argument?

```
let ss1: Subspace = (1,).into();
let ss2: Subspace = 1.into();
assert_eq!(ss1.bytes(), ss1.bytes()); // true
```

At first time thought it is a kind of bug, but it seems quite convenient and clean in subspace case. @bluejekyll Could you share your opinion on this?

rushmorem:
I doubt that it's a bug. It looks like intended behaviour to me because we actually have a test for this particular behaviour (`tuple::tests::test_eq`):-

```rust
    #[test]
    fn test_eq() {
        // ...
        assert_eq!("string".encode_to_vec(), ("string",).encode_to_vec());
    }
```

Personally, I like it and I can't think of any problems that might be caused by it.

bluejekyll:
`assert_eq!("string".encode_to_vec(), ("string",).encode_to_vec());`

Hm, perhaps that test was bad to put in... it might be wrong. I was double checking some assumptions about tuples and bare values in the way our library was working. It's not clear to me that is correct.

rushmorem:
> Hm, perhaps that test was bad to put in... it might be wrong. I was double checking some assumptions about tuples and bare values in the way our library was working. It's not clear to me that is correct.

In that case let me revert my last commit just to be on the safe side.

bluejekyll:
See #83 for the inner tuple fix...

rushmorem:
@bluejekyll I believe this is now ready to merge.

bluejekyll:
Thanks for the PR, this is a good cleanup.

We need to validate this, but as far as I can tell `(1,).encode() == 1.encode()`, but we really need to validate that. Especially b/c I'm fairly certain that `(1, (1,)).encode() != (1, 1).encode()` and that's a little awkward...

rushmorem:
> Thanks for the PR

No problem :)

> We need to validate this, but ...

Thanks, I have updated the tracking issue.

PR #80, opened 2018-05-07
reduce allocations when decoding

rushmorem:


yjh0502:
Looks good to me. Thanks!

Issue #82, opened 2018-05-07
Is it okay to allow converting an element to a single-element tuple implicitly?


rushmorem:
https://github.com/bluejekyll/foundationdb-rs/pull/78#issuecomment-387089309

rushmorem:
@bluejekyll in https://github.com/bluejekyll/foundationdb-rs/pull/78#issuecomment-387230049

> We need to validate this, but as far as I can tell `(1,).encode() == 1.encode()`, but we really need to validate that. Especially b/c I'm fairly certain that `(1, (1,)).encode() != (1, 1).encode()` and that's a little awkward...

rushmorem:
I like that. However, I see a lot of examples with single element tuples in FDB documentation. I'm not sure whether they are supposed to be encoded differently from non-tuple values.

bluejekyll:
Ok, based on this comment, with python generated binary slices: https://github.com/bluejekyll/foundationdb-rs/pull/77#discussion_r186330604

I added two test cases too #83: 

https://github.com/bluejekyll/foundationdb-rs/blob/bbf3241f634ce8d91c961db5712b9daa0a7e9d1c/foundationdb/src/tuple/mod.rs#L271-L277

and

https://github.com/bluejekyll/foundationdb-rs/blob/bbf3241f634ce8d91c961db5712b9daa0a7e9d1c/foundationdb/src/tuple/mod.rs#L303-L309

Which. If you look at those encodings in python, and from that comment, there is no nested tuple markers for the top level tuple. So this is correct at that point.

I think we still need to verify what happens with the single inner tuple...

yjh0502:
Here's some nested tuple examples,

```
>>> [ord(x) for x in fdb.tuple.pack((1,(1,)))]
[21, 1, 5, 21, 1, 0]
>>> [ord(x) for x in fdb.tuple.pack( (1,(1,(1,))) )]
[21, 1, 5, 21, 1, 5, 21, 1, 0, 0]
>>> [ord(x) for x in fdb.tuple.pack( (1,(1,),(1,)) )]
[21, 1, 5, 21, 1, 0, 5, 21, 1, 0]
>>>
```

bluejekyll:
So I guess for maximal compatibility we’ll need single tuple. Oh well, I think the new depth feature on tuple encoding will deal with this fine. I’ll add single tuple back.

PR #83, opened 2018-05-07
track when inside tuples during encoding

bluejekyll:
Fixes: #49 

This resurrects and corrects the tests for recursive tuples. It's not a super elegant solution, but I think is correct.

- adds `tuple_depth: usize` to encode and decode
- adds `Option` encode/decode with tuple encoding edge case test

### DIFF
--- a/foundationdb/examples/class-scheduling.rs
+++ b/foundationdb/examples/class-scheduling.rs
@@ -78,8 +78,8 @@ fn init_classes(trx: &Transaction, all_classes: &[String]) {
 
 fn init(db: &Database, all_classes: &[String]) {
     let trx = db.create_trx().expect("could not create transaction");
-    trx.clear_subspace_range(&"attends");
-    trx.clear_subspace_range(&"class");
+    trx.clear_subspace_range("attends");
+    trx.clear_subspace_range("class");
     init_classes(&trx, all_classes);
 
     trx.commit().wait().expect("failed to initialize data");
@@ -98,8 +98,7 @@ fn get_available_classes(db: &Database) -> Vec<String> {
                 let count = i64::try_from(key_value.value()).expect("failed to decode count");
 
                 if count > 0 {
-                    let class =
-                        String::try_from(key_value.key()).expect("failed to decode class");
+                    let class = String::try_from(key_value.key()).expect("failed to decode class");
                     available_classes.push(class);
                 }
             }
--- a/foundationdb/examples/class-scheduling.rs
+++ b/foundationdb/examples/class-scheduling.rs
@@ -70,7 +70,7 @@ fn all_classes() -> Vec<String> {
 }
 
 fn init_classes(trx: &Transaction, all_classes: &[String]) {
-    let class_subspace = Subspace::from(&"class");
+    let class_subspace = Subspace::from("class");
     for class in all_classes {
         trx.set(&class_subspace.pack(class), &100_i64.to_vec());
     }
--- a/foundationdb/examples/class-scheduling.rs
+++ b/foundationdb/examples/class-scheduling.rs
@@ -70,7 +70,7 @@ fn all_classes() -> Vec<String> {
 }
 
 fn init_classes(trx: &Transaction, all_classes: &[String]) {
-    let class_subspace = Subspace::from("class");
+    let class_subspace = Subspace::from(&"class");
     for class in all_classes {
         trx.set(&class_subspace.pack(class), &100_i64.to_vec());
     }
@@ -78,8 +78,8 @@ fn init_classes(trx: &Transaction, all_classes: &[String]) {
 
 fn init(db: &Database, all_classes: &[String]) {
     let trx = db.create_trx().expect("could not create transaction");
-    trx.clear_subspace_range(("attends",));
-    trx.clear_subspace_range(("class",));
+    trx.clear_subspace_range(&"attends");
+    trx.clear_subspace_range(&"class");
     init_classes(&trx, all_classes);
 
     trx.commit().wait().expect("failed to initialize data");
--- a/foundationdb/src/tuple/element.rs
+++ b/foundationdb/src/tuple/element.rs
@@ -678,7 +678,7 @@ mod tests {
 
     #[test]
     fn test_option() {
-        assert_eq!(&Some(42_i64).encode_to_vec(), &[21, 42]);
-        assert_eq!(&None::<i64>.encode_to_vec(), &[0]);
+        assert_eq!(&Some(42_i64).to_vec(), &[21, 42]);
+        assert_eq!(&None::<i64>.to_vec(), &[0]);
     }
 }
--- a/foundationdb/src/tuple/element.rs
+++ b/foundationdb/src/tuple/element.rs
@@ -3,7 +3,7 @@ use std::{self, io::Write};
 use uuid::Uuid;
 
 use byteorder::{self, ByteOrder};
-use tuple::{Decode, Encode, Error, Result, Tuple};
+use tuple::{Decode, Encode, Error, Result, Tuple, TupleDepth};
 
 /// Various tuple types
 pub(super) const NIL: u8 = 0x00;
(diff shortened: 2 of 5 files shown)


foundationdb/src/tuple/element.rs
```diff
@@ -256,7 +256,8 @@ impl Decode for String {
 }
 
 impl Encode for Vec<Element> {
-    fn encode<W: Write>(&self, w: &mut W) -> std::io::Result<()> {
+    fn encode<W: Write>(&self, w: &mut W, _in_tuple: bool) -> std::io::Result<()> {
+        // TODO: should this only write the Nested markers in the case of in_tuple?
```
bluejekyll: I think we only want to write the Nested values in the actual nested case, right? I didn't change this, but it seems like this is wrong, and should follow the same patter in the generated tuple macro.

bluejekyll:
FYI, I just added a change to this PR to drop the single tuple variant

Fixes: #82

foundationdb/src/tuple/mod.rs
```diff
@@ -139,7 +163,6 @@ macro_rules! tuple_impls {
 }
 
 tuple_impls! {
-    1 => (0 T0)
```
bluejekyll: By removing this, I think it clarifies single tuple vs. single element in a lot of contexts.

bluejekyll:
Ok, I think this is ready to land.

bluejekyll:
Ok, I've added unary tuple support back, in accordance with #82

bluejekyll:
@yjh0502 and @rushmorem I think after we land this, we're good for the next release.

foundationdb/src/tuple/element.rs
```diff
@@ -459,21 +461,63 @@ impl Decode for i64 {
...
+    T: Decode,
+{
+    fn decode(buf: &[u8], tuple_depth: usize) -> Result<(Self, usize)> {
+        match buf[0] {
+            NIL => {
+                let buf = &buf[1..];
+
+                // custom escape markers are only needed in Nested tuples...
+                if tuple_depth > 1 {
+                    ESCAPE.expect(buf[0])?;
```
yjh0502: We need a bound check here, too.

foundationdb/src/tuple/mod.rs
```diff
@@ -189,13 +212,13 @@ mod tests {
 
     #[test]
     fn test_malformed_int() {
-        assert!(Tuple::decode(&[21, 0]).is_ok());
-        assert!(Tuple::decode(&[22, 0]).is_err());
-        assert!(Tuple::decode(&[22, 0, 0]).is_ok());
+        assert!(Tuple::decode(&[21, 0], 0).is_ok());
```
yjh0502: Could we use `try_from` instead of `decode`, as following tests do?

foundationdb/src/tuple/element.rs
```diff
@@ -459,21 +461,63 @@ impl Decode for i64 {
     }
 }
 
+impl<T> Encode for Option<T>
```
yjh0502: BTW It looks good!

foundationdb/src/tuple/mod.rs
bluejekyll: sure, I can make that change.

foundationdb/examples/class-scheduling.rs
```diff
@@ -70,16 +70,16 @@ fn all_classes() -> Vec<String> {
 }
 
 fn init_classes(trx: &Transaction, all_classes: &[String]) {
-    let class_subspace = Subspace::from("class");
+    let class_subspace = Subspace::from(&"class");
```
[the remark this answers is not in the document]
bluejekyll: I needed to add that. `Subspace::from` may not be properly specified, I'll look into it.

foundationdb/src/tuple/element.rs
bluejekyll: thanks!

foundationdb/examples/class-scheduling.rs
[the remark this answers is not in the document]
yjh0502: Maybe it works only on nightly compiler.. I didn't thought it might not work on stable because my default toolchain is a nightly.. Please ignore this comment, sorry for the noise.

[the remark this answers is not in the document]
bluejekyll: Ok, I just double checked... the CLI cargo build works fine. Apparently it's the VSCode plugin that is reporting an error when it is not. Odd.

bluejekyll:
Thanks for the review. I added the bounds checks... but oddly I couldn't actually figure out a breaking test for them, so there may not have been an issue? Anyway, the new variant is far safer.

bluejekyll:
@rushmorem see https://github.com/bluejekyll/foundationdb-rs/pull/83/commits/83765ffecc72445c9382ea9fa1345b5a526b54f7

This introduces `TupleDepth` with a private constructor and then the `encode_to` and `decode_from` methods. External libraries won't be able to construct `TupleDepth`. I suppose we could even make it more private if we want, so that it can only be constructed in those top level calls. If you look at the bindingstester binary, that has examples of the new interface. Thoughts?

bluejekyll:
@yjh0502 want to give this a final once over?

foundationdb/src/tuple/mod.rs
```diff
@@ -35,6 +36,25 @@ pub enum Error {
     FromUtf8Error(FromUtf8Error),
 }
 
+/// Tracks the depth of a Tuple decoding chain
+#[derive(Copy, Clone)]
+pub struct TupleDepth(usize);
+
+impl TupleDepth {
+    pub(crate) fn new() -> Self {
```
rushmorem: This is just brilliant. I love it!

bluejekyll: oops, that pub(crate) isn't necessary... let me remove that.

foundationdb/src/tuple/mod.rs
```diff
@@ -35,6 +36,25 @@ pub enum Error {
...
+        TupleDepth(0)
+    }
+
+    fn increment(&self) -> Self {
```
rushmorem: Doesn't this need to be public?

foundationdb/src/tuple/mod.rs
```diff
@@ -65,28 +85,40 @@ impl DerefMut for Tuple {
...
     /// Encodes this tuple/elemnt into the associated Write
-    fn encode<W: Write>(&self, _w: &mut W) -> std::io::Result<()>;
+    fn encode<W: Write>(&self, _w: &mut W, tuple_depth: TupleDepth) -> std::io::Result<()>;
+
+    /// Encodes this tuple/elemnt into the associated Write
+    fn encode_to<W: Write>(&self, w: &mut W) -> std::io::Result<()> {
+        self.encode(w, TupleDepth::new())
+    }
+
     /// Encodes this tuple/elemnt into a new Vec
```
rushmorem: While you are at it, would you be so kind to fix these typos? Notice `elemnt` in this comment and the other 2 before it.

foundationdb/src/tuple/mod.rs
```diff
@@ -35,6 +36,25 @@ pub enum Error {
...
+/// Tracks the depth of a Tuple decoding chain
+#[derive(Copy, Clone)]
+pub struct TupleDepth(usize);
+
+impl TupleDepth {
+    fn new() -> Self {
+        TupleDepth(0)
+    }
+
+    fn increment(&self) -> Self {
```
rushmorem: I don't know what happened to my earlier comment but doesn't `increment` need to be public? Won't someone implementing the traits externally need access to it?

foundationdb/src/tuple/mod.rs
```diff
@@ -224,11 +278,115 @@ mod tests {
...
-            "string".to_vec(),
-            "string".to_string().to_vec()
+            &("one", ("two", 42, ("three", 33))).to_vec(),
+            &[
+                2, 111, 110, 101, 0, 5, 2, 116, 119, 111, 0, 21, 42, 5, 2, 116, 104, 114, 101, 101,
+                0, 21, 33, 0, 0,
+            ]
         );
 
-        assert_eq!("string".to_vec(), ("string",).to_vec());
```
rushmorem: Isn't this test still relevant, and even important?

bluejekyll: I added a separate set of tests for nested tuples. I can add this back.

foundationdb/src/tuple/mod.rs
bluejekyll: Yes. You are correct. I will edit.

foundationdb/src/tuple/mod.rs
bluejekyll: Yep.

foundationdb/src/tuple/mod.rs
rushmorem: Yes I noticed, but I think this one is unique in that I haven't seen any other test comparing a single element to a tuple with a single element.

bluejekyll: technically there is this one :`assert_eq!(&(1,).to_vec(), &1.to_vec());` but it's ok, I've added this back.

PR #88, opened 2018-05-12
Add `futures` dependency to example.

eonil:
The example doesn't work out-of-the-box without the dependency.
It took several hours to figure out what's wrong for me.
This can help other Rust newbies like me...

### DIFF
--- a/foundationdb/README.md
+++ b/foundationdb/README.md
@@ -29,6 +29,7 @@ $> sudo installer -pkg FoundationDB-5.1.5.pkg -target /
 ```toml
 [dependencies]
 foundationdb = "*"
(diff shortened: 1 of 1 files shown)


bluejekyll:
Thanks for the contribution! Glad you were able to work through the issue.

tirkarthi:
Another place will be https://github.com/bluejekyll/foundationdb-rs/blob/42782060a5352d338c90462d97b32cae0182c166/foundationdb/src/lib.rs#L37 

I used the generated [docs](https://docs.rs/foundationdb/0.2.0/foundationdb/) instead of README and faced the same error.

Issue #93, opened 2018-06-08
Transaction get, get_range return types


Speedy37:
Hi,

I was wondering about the Transaction api:

Why is `Transaction.get` future not directly returning `Option<Vec<u8>>` ?

I propose:

```rust
fn get(&self, key: &[u8], snapshot: bool) -> Future<Option<&[u8]>>;
fn get_range(&self, opt: RangeOption) -> Stream<KeyValue>;
fn get_key(&self, selector: KeySelector, snapshot: bool) -> Future<&[u8]>;
```

What do you think ?

yjh0502:
There are several considerations on current API.

 - You might want to get additional information about the query after fetching the value. For example, https://docs.rs/foundationdb/0.1.0/foundationdb/transaction/struct.Transaction.html#method.committed_version

 - Current API does not copy a value from fdb api, and a caller can copy it if they need to. If we return Vec, the value will be always copied.
 - It's common to call multiple read commands, or call write command after read. With current API, caller can get a `Transaction` which can be used to continue operation with same transaction. If API returns value only, caller need to clone the transactions and carry it around closures.

Speedy37:
Point 1 & 3 are about retaining the transaction object: there is one indirection just for that case, while the transaction object is easy to clone. I don't see this as an advantage.

Point 2, I meant to return `&[u8]` (probably require some lifetime declaration with it: as long as the Future is alive, &[u8] is alive).

Speedy37:
For `get_range`, I can imagine a need to be able to manually access "bunchs" of key/value but I think the most useful case is just: iterate and at any point in the iteration being able to "generate" a "next" option for future calls to `get_range` (for pagination).

bluejekyll:
So basically you want to drop the `Transaction` object as part of the `GetResult` future result from the `get`s, right? And then the user would need to clone `Transaction` if they plan to use this in any other lambdas or future combinators, right?

Another option would be to make `GetResult` (and the others) simply implement `Deref<Result<Option<&[u8]>>` and then it would appear to have the interface you're asking for, right? That would be the most minimal change and leave the chaining in place that @yjh0502 likes.

Speedy37:
Right. It's a tradeoff I would take. I'll make a real life example with both cases.

Actually I prefer `value()` to deref, it's just to remove the "double Result" check (by the future and inside then).

bluejekyll:
[...]

PR #94, opened 2018-06-08
change `TrxGet::value` to return `Option<&[u8]>`

yjh0502:
Related to: #93

### DIFF
--- a/foundationdb/src/transaction.rs
+++ b/foundationdb/src/transaction.rs
@@ -646,16 +646,11 @@ impl Future for TrxGet {
     type Error = Error;
 
     fn poll(&mut self) -> std::result::Result<Async<Self::Item>, Self::Error> {
-        let res = match self.inner.poll()? {
-            Async::Ready((trx, inner)) => {
-                // check if a future resolves to value type
-                inner.get_value()?;
+        let (trx, inner) = try_ready!(self.inner.poll());
+        // check if a future resolves to value type
+        inner.get_value()?;
 
-                Async::Ready(GetResult { trx, inner })
-            }
-            Async::NotReady => Async::NotReady,
-        };
-        Ok(res)
+        Ok(Async::Ready(GetResult { trx, inner }))
     }
 }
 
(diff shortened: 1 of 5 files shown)


foundationdb/src/transaction.rs
```diff
@@ -630,8 +630,10 @@ impl GetResult {
     }
 
     /// Returns the values associated with this get
-    pub fn value(&self) -> Result<Option<&[u8]>> {
-        self.inner.get_value()
+    pub fn value(&self) -> Option<&[u8]> {
+        self.inner
+            .get_value()
+            .expect("inner should resolve into value")
```
bluejekyll: What are the cases where this could fail? It's a binary slice, so I'm thinking it would be a problem with the underlying C api, right?

foundationdb/src/transaction.rs
```diff
@@ -831,11 +827,8 @@ impl Future for TrxCommit {
     type Error = Error;
 
     fn poll(&mut self) -> std::result::Result<Async<Self::Item>, Self::Error> {
-        match self.inner.poll() {
-            Ok(Async::Ready((trx, _res))) => Ok(Async::Ready(trx)),
-            Ok(Async::NotReady) => Ok(Async::NotReady),
-            Err(e) => Err(e),
-        }
+        let (trx, _res) = try_ready!(self.inner.poll());
```
bluejekyll: these are nice cleanups!

foundationdb/src/transaction.rs
yjh0502: Yes. fdb-c `fdb_future_get_value` might return error when the future returns a value other then value, for example, key or keyvalues.

https://github.com/bluejekyll/foundationdb-rs/blob/master/foundationdb/src/future.rs#L192-L199

To make sure, we check if it resolves to value type before returning `TrxGet`, so I think those `expect` should not be reachable.

https://github.com/bluejekyll/foundationdb-rs/blob/master/foundationdb/src/future.rs#L192-L199

bluejekyll:
Merging. This looks good. Sorry for the delay.

PR #97, opened 2018-11-16
WIP: auto-generate all code at build time

thoughtpolice:
**Problem**: I wanted to use this library on Windows with FoundationDB 5.2, but I noticed that all of the code for the options and `-sys` bindings were generated offline, i.e. not at compile time. This means setting a `runtime_version` of `520` failed when I tried it (even though I was using the 5.2 client libraries.)

**Fix**: This PR changes the build system so that all code related to `fdb.options` and `foundationdb-sys` are generated at build time:

  - `foundationdb-sys`: now has a set of features `fdb-5_1`, `fdb-5_2` and `fdb-6_0` which control what version of `FDB_API_VERSION` to set. (This defaults to `fdb-5_2` for now.)
  - `foundationdb-sys`: the `wrapper.h` header file is now generated "on demand" by `build.rs` by simply writing it out. This allows end-users to control the version, by way of the introduced `#cfg` features (because previously there was no way to control the hard-coded `#define` macro in the `.h` file.)
  - `foundationdb-gen` now reads `fdb.options` directly out of the filesystem, rather than using an included version in the source code. This is because *all* C-API client libraries ship `fdb.options`, right next to the C headers, precisely so that it can be used for 3rd-party binding generators. Unfortunately, this means replicating the `INCLUDE_PATH` hack that was present in `build.rs` for foundationdb -- the paths are all hardcoded (although this isn't strictly *worse* than before.)
  - `foundationdb-gen` had all of the code for generating options moved into a library, in the top-level `foundationdb_gen` crate/module. This module emits one public function, `foundationdb_gen::emit`, which gives back a `Result<String>` containing all the Rust code for `options.rs`.
  - `foundationdb-gen` still has an executable that simply prints out the code to `stdout`, so you can test it.
  - `foundationdb-gen` was also lightly tweaked to support FoundationDB 5.2+ (there was a new option added that caused slightly incorrect code to be generated due to the snakecase library; `AppendIfFit` should have been `AppendIfFits`, and fixing this required some minor touchups.)
  - `foundationdb` now uses the `foundationdb-gen` library as a `build-dependency`, and has a new `build.rs` file which writes out `options.rs` on demand into `OUT_DIR`. This is read from `options.rs` in the source code, directly.

The now-configurable `FDB_API_VERSION` flag means that clients will need to pick the right API version they want to support via Cargo, e.g. 

```
[dependencies]
foundationdb = "..."

[dependencies.foundationdb-sys]
default-features = false
default-features = [ "fdb-6_0" ]
```

This is a fairly large change in terms of diffstat, but the overall core changes are very small. I think it still needs some work though; in particular, new docs for users, I suppose.


bluejekyll:
Quick feedback. I’m personally open to this change, but there is a bit of history here. Originally we were generating all the bindings at compile time, but this caused some stability issues between Linux and macOS, as I remember. If we can stabalize the build across all threee platforms then I think this is fine.

We will need to publish the `foundationdb-gen` crate, which up to now hasn’t been required. Just something to note for the next time we publish.

thoughtpolice:
I've tried bumping the CI builds to use FoundationDB 5.2.x, as well as removing `--all-features` from the CI build for `foundationdb-sys` (since otherwise it implies *all* API versions and there is no way to make a Cargo feature mutually exclusive.)

I notice that [the FoundationDB website](https://www.foundationdb.org/download/) has also released download links to 6.0.15 -- I knew the 6.0.x release was coming soon, but haven't seen any announcement.

Would it perhaps be better to go ahead and move the default features to `fdb-6_0` and install the 6.x packages, do you think?

thoughtpolice:
It seems like the FoundationDB 5.2.x packages have newer dependencies than Ubuntu Trusty (14.04) would allow, so I've bumped the CI builds to use Xenial (16.04) instead.

I have also pushed a commit moving to 6.0 (and feature `fdb-6_0`) by default. Let's see how it goes.

bluejekyll:
@yjh0502, I had to escape the smoke of SF, I might not be able to review this in detail. Would you be able to?

foundationdb/build.rs
```diff
@@ -0,0 +1,17 @@
+extern crate foundationdb_gen;
+
+use std::env;
+use std::fs::File;
+use std::io::prelude::*;
+use std::path::PathBuf;
+
+fn main() {
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
```
yjh0502: Could we use `expect` instead of unwrap? It will make troubleshooting much easier.

foundationdb-gen/src/bin/foundationdb-options-gen.rs
```diff
@@ -1,384 +1,5 @@
...
-        println!("{}", scope.gen_ty());
-        println!("{}", scope.gen_impl());
-    }
-    Ok(())
-}
+extern crate foundationdb_gen;
 
 fn main() {
-    run().expect("failed to run");
+    println!("{}", foundationdb_gen::emit().unwrap());
```
yjh0502: Could we add an error message with `expect`?

yjh0502:
It seems that there was a problem on my environment.. It works on a new shell (I don't know why). Thanks!

thoughtpolice:
I have fixed all of the calls to `.unwrap` that I added/modified. Should be good now, but needs docs still for the feature flags.

PR #99, opened 2018-11-21
Cleanup the fdb version features from primary crate

bluejekyll:
There was no decent way to select FDB versions from the primary crate.
  this disables all features in the sys crate from the fdb crate,
  and clones all the feature names to make it possible to enable
  each as necessary.


bluejekyll:
@yjh0502 any concerns with me merging this?

yjh0502:
Oh, sorry for confusion. I understood your point, and the solution looks fine!

PR #143, opened 2019-11-02
async/await: complete rewrite of foundationdb

### LINKED ISSUE #111: Support stablization track futures API
The API for rust futures has changed a bit since the `futures 0.1` crate, and the current version of the API is on track for stablization.

[futures-preview](https://crates.io/crates/futures-preview/0.3.0-alpha.12) exposes the new API, although internally its just re-exposing std::future. It'd be rad to support this new API in foundationdb-rs

### LINKED ISSUE #117: Support for u64 elements in tuples
According to Apple's [documentation](https://github.com/apple/foundationdb/blob/master/design/tuple.md) the tuple layer should be able to support u64 elements:

> 0x1c is an 8 byte positive number

Unless i am mistaken the Rust client currently does not support that.

### LINKED ISSUE #123: should `get_range` return a `KeyValue` stream?
`Transaction#get_ranges` is a little tricky to use. It returns a `Stream<GetRangeResult>`, where `GetRangeResult` itself contains a `KeyValues` which derefs into a `[KeyValue]`. This means an extra layer of iteration to loop through all values in a range.

It'd be nice to offer a stream to the underlying `KeyValue` entries directly, which is something the other higher-level bindings do (links below!).

My proposal would be to:
* remove the current flavor of `get_range` as a public function (it forces you to think about iteration count, which isn't intuitive + is only relevant with certain streaming modes)
* make the current `get_ranges` return `Stream<KeyValue>`
* rename `get_ranges` to `get_range`

I think this would bring the binding into closer alignment with the others for range reads.

[1] : Java returns a `Iterable<KeyValue>` (https://apple.github.io/foundationdb/javadoc/index.html)
[2] : Go returns a `RangeResult` which can either be used to pull all values into a slice, or act as an iterator(https://godoc.org/github.com/apple/foundationdb/bindings/go/src/fdb#Transaction.GetRange)
[3] : Python returns a generator over the KVs (https://github.com/apple/foundationdb/blob/master/bindings/python/fdb/impl.py#L327 / https://apple.github.io/foundationdb/api-python.html)

### LINKED ISSUE #124: Supporting 6.1 API
FDB 6.1.8 is the current stable release, and the 6.1 API has some backward incompatible changes. Here are the details on the changes:

* https://github.com/apple/foundationdb/pull/942
* https://forums.foundationdb.org/t/upcoming-c-api-changes-planned-for-version-6-1/1058

I took a stab at making a PR to support the `610` API version, but it will require either a) breaking backward compatibility or b) wrapping code with `#[cfg(feature = "fdb-6_1")]` blocks all over the place.

Since it's a bit messy, I figured I'd start a discussion first before moving forward. What do people think the best way to move forward is? My preference is to prevent breaking backward compatibility at the expense of introducing some more complexity.

In the meantime, I have a small patch here that uses the `600` API version with the 6.1.x client libraries: https://github.com/brndnmtthws/foundationdb-rs/commit/ee95ad4f1efc1a6050a3f3f8e9974612b545e2b5

### LINKED ISSUE #125: Errors in db.transact() closure are not retried
Hi guys,

I'm seeing an issue where some transactions run with db.transact() are timing out with "Error { kind: \n\nFoundationDB error(1009): Request for future version, should_retry: true }" after 1 second, when a storage server gets a little lagged. I'm puzzled as to why the transaction is not retried, as I'm not setting any transaction options, and I gather there are no timeouts by default. I don't have any experience using foundationdb in other language bindings, so I'm not sure if this is an issue with the rust bindings or in foundationdb itself. If it's the latter, please let me know and I'll take the issue up on the foundationdb forums.

### LINKED ISSUE #128: implement versionstamp bindings for tuple layer
Versionstamps are a compelling feature that tie in closely with the tuple layer, and are the main thing currently missing from it right now.

They have a bit of a weird API, where you encode an incomplete versionstamp to a tuple, pack the tuple with a special offset (w/ different values depending on FDB client version), write the resulting bytes with the `SET_VERSIONSTAMP_(KEY|VALUE)` atomic operations, and then FDB magically replaces the incomplete-versionstamp portion of the tuple with the actual commit-time versionstamp when it writes to disk.

They were recently added to the golang bindings: https://github.com/apple/foundationdb/pull/1187/files which makes for a nice distillation of what they need to work. 

That said: they do puncture a bit of the clean implementation of tuples we have now, since they add some new invariants:

* vanilla packing of a tuple cannot contain an incomplete versionstamp, otherwise it fails
* versionstamp-packing a tuple must contain _exactly_ one incomplete versionstamp, otherwise it fails

### LINKED ISSUE #130: rename `Encode::to_vec` --> `Encode::pack` ?
This is minor, but one thing I've noticed hopping between language bindings is that all of the various FDB tuple layers use `pack` and `pack_with_versionstamp` as the verbs for turning a tuple into bytes, and currently we use `to_vec` (we don't have a with-versionstamp variant yet). `to_vec` seems more customary around Rust libs, but I'm wondering if it makes sense to rename for consistency w/ the other FDB bindings, or have both.

### LINKED ISSUE #132: Options.rs singularization in FDB-gen for FDB 6.1
fdb-client 6.1.8 installed, can use 6.0 bindings. Had to move forward from release-3.0
to master since I am running into the issue fixed by both #120 #121

Issue occurs on attempting to compile foundationdb crate
in bindings.rs
```
pub const FDBTransactionOption_FDB_TR_OPTION_USE_PROVISIONAL_PROXIES: FDBTransactionOption = 711;

```

Error:
```

error[E0425]: cannot find value `FDBTransactionOption_FDB_TR_OPTION_USE_PROVISIONAL_PROXY` in module `fdb`
   --> /home/testuser/rust/test-project/target/debug/build/foundationdb-c9fdf7023e981c03/out/options.rs:368:48
    |
368 | TransactionOption::UseProvisionalProxy => fdb::FDBTransactionOption_FDB_TR_OPTION_USE_PROVISIONAL_PROXY,
    |                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: a constant with a similar name exists: `FDBTransactionOption_FDB_TR_OPTION_USE_PROVISIONAL_PROXIES`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0425`.
error: Could not compile `foundationdb`.

To learn more, run the command again with --verbose.

```

Seems like doing this fixes that issue
Assignee
```
if opt.name == "AppendIfFit" {
    opt.name = String::from("AppendIfFits");
} else if opt.name == "UseProvisionalProxy" {
    opt.name = String::from("UseProvisionalProxies");
};
```
Places to be edited for that to function: lines 64, 77, and 283 

Reason for issue:

inflector::cases::screamingsnakecase singularizes plural words by default. I wonder if there's a way to turn that off.

Edit: Turns out it's not screamingsnakecase, but classcase that does it, changing from classcase to pascalcase solves the issue for current and future possible spelling issues since pascalscase does not singularize by default.

### LINKED ISSUE #135: Async/await and FDB 6.1 support
I've ported this library to edition 2018, futures 0.3 and foundationdb 6.1 as a learning exercise. All the tests pass, and the bindingtester does not report any problems.

As the async changes will require a version number bump, I thought it made sense to make the breaking changes required to support FDB 6.1 at the same time, instead of going to the extra trouble of trying to support both APIs at once (https://github.com/bluejekyll/foundationdb-rs/issues/124)

If you'd like me to send a pull request let me know. In the mean time users can try it out on my branch: https://github.com/dae/foundationdb-rs/tree/futures0.3-rebased

Speedy37:
I don't know if this commit can be the future of this project or if I should publish it on my own with another name. I changed the API and logic quite a lot.

The most important change to me is how transactions are shared across threads. They can't be cloned anymore, so you share them by reference, Transaction implements Send/Sync, but commit/cancel/reset api requires mutable access to a Transaction. This protect against undefined behavior that was previously possible (cancel/reset) data races.

 - no more indirection within FdbFuture
 - no race condition/undefined behavior possible (Transaction can't be cloned anymore) and commit, cancel, reset are protected
 - fdb api 610+
 - option generation is now indented and the code is simpler
 - RangeOption and KeySelector can be either be Owned or Borrowed
 - KeySelector offset can be negative (there is a test of this in the binding checker, this was not found due to casting luck)
 - some int options can be negative
 - fix boot api safety (undefined behavior was possible)
 - simple boot process
 - use rustc stable in 1.39 async/await feature
 - foundationdb 510, 520, 600 support with common Database::new_compat api

I'm marking this as WIP as I still need to polish/document it.
Here is a small/incomplete TODO of what I want to do:

 - [x] check every single integer casts
 - [x] threaded bindingtester
 - [x] write doc
 - [x] continuous integration
 - [ ] directory / hca

I've not focused much on the tuple/subspace/directory API as I'm not using them in production. I built a strictly typed API that enforce database layout.

Fixes #111
Fixes #117
Fixes #123
Fixes #124
Fixes #125
Fixes #128
Fixes #130
Fixes #132
Fixes #135


bluejekyll:
@brndnmtthws, thoughts? I currently don’t have skin in the game, but this is a significant rewrite and change in direction.

Speedy37:
Well serde was a bad idea to implement the tuple layer, I'm converting it back.

Speedy37:
This latest version passes the following bindingtester tests 100 times (might still miss some low occurrences bugs).

```sh
./bindingtester.py --num-ops 1000 --test-name api --api-version 610
./bindingtester.py --num-ops 1000 --concurrency 5 --test-name api --api-version 610
./bindingtester.py --test-name scripted
```

Most bugs were in the tuple pack/unpack code and in the bindingtester code itself.
It was quite hard to keep the state of the bindingtester ok due to the new safety measures of the new api. I had to simulated get committed version on a non committed transaction for example (last commit).

dae:
I haven't reviewed the code properly, but have no objections to my PR being discarded if this is a better long term solution.

@Speedy37 : Is the retry/timeout code in transact() necessary? I was under the impression FDB handles that itself when on_error is called.
https://apple.github.io/foundationdb/api-c.html#c.fdb_transaction_on_error

Speedy37:
I dropped the serde approach to ser/deserialization as some types (Versionstamp, Uuid) were quite hard to support. It might be possible to add it back as `impl TuplePack/TupleUnPack for T where T: Serialize/Deserialize` which would make "named" tuple possible by using structs.

I had like to not fork too. I'm also open to suggestion and would recommend not merging this until rustc 1.39 is out. I still need to finish porting my app whole stack to this version to validate the usability of the API.

In the current state, I'm not sure about the signature of Database::transact.

Fdb does handle the retry count itself if specified, so this limit is a duplicate.

Speedy37:
Ok this is only missing the directory layer API now.

Speedy37:
Here is a link to the latest generated documentation: https://clikengo.github.io/foundationdb-rs/foundationdb/index.html

Speedy37:
Ok for now, the biggest downside of this version is that `Database::transact` is not the easiest to use.
For me there is two way to approach how to use foundationdb with this branch:

 - share a transaction reference, this is easy, but might require some lifetimes hints to the compiler which might be hard for new to rust users. On the other side, doing it this way, force you to prove to the compiler that what you are doing makes sense.
 - share a reference counted transaction (either Arc or Rc). It's is easy to do, but, also there is no compiler proof at commit/abort/reset point that there is only one reference to the transaction alive. It's something that is done at runtime.

josephg:
This is awesome - thankyou so much for getting this working! 😍

josephg:
FWIW I think sharing a transaction reference is the right API - getting static lifetime checks is good rust code, even if it means users need lifetime qualifiers for the transaction object.

pH14:
🎉 this is such a huge step forward, amazing work

PR #145, opened 2019-11-06
Mark FdbFuture as Send + Sync

Daniel-B-Smith:
The primary motivation here is to make the transaction commit future compatible with Tokio. Tokio requires all futures spawned on its executor to be `Send`. I added the `Sync` as well since the two tend to go together.

From reading the docs and inspecting the code, I see no reason that the future cannot be safely sent between threads. The only method on the struct can only be called from one thread at a time, so the struct should be implicitly `Sync`.

The dependency on `static_assertions` is not strictly required, but it does make asserting the trait implementations I care about easier.


Daniel-B-Smith:
Thanks!

PR #152, opened 2019-11-20
Make the various result types Send/Sync

Daniel-B-Smith:
Related to #145. This change marks the FDBResult type as Send/Sync and asserts that all of the transaction result types are thus Send/Sync. The FDB C API is thread safe, so this change should be safe.

### DIFF
--- a/foundationdb/src/future.rs
+++ b/foundationdb/src/future.rs
@@ -149,6 +149,9 @@ impl Drop for FdbFutureResult {
     }
 }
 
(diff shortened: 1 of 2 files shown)


Speedy37:
I'm guessing you are using the master branch? Be aware, I'm gonna merge the async branch into master very soon. So this will be merged to the 0.3 branch only.

Daniel-B-Smith:
Thanks for the heads up! I am indeed using the master branch and intend to switch over once the async branch is merged.

PR #155, opened 2019-11-26
Fix Element::Bool always false

### LINKED ISSUE #150: Bool element always false
all bool elements are being stored as false.

Aioxas:
* Resolves #150


Speedy37:
Thanks

Issue #170, opened 2020-01-23
`fdb_stop_network` MUST be called before the program exits.


Speedy37:
Sometimes the tokio test fails:

```
error: test failed, to rerun pass '--test tokio'
##[error]test failed, to rerun pass '--test tokio'
Caused by:
  process didn't exit successfully: `/home/runner/work/foundationdb-rs/foundationdb-rs/target/debug/deps/tokio-c51e2bb533a18041` (signal: 6, SIGABRT: process abort signal)
##[error]The process '/usr/share/rust/.cargo/bin/cargo' failed with exit code 101
##[error]Node run failed with exit code 1
```

Speedy37:
This was with v0.2.10.
Seems like recent builds (v0.2.11) aren't affected.

Speedy37:
Nevermind, still failing sometimes: https://github.com/Clikengo/foundationdb-rs/pull/173/checks?check_run_id=422465156

Speedy37:
This is probably caused due to this foundation spec:

`fdb_stop_network`:
Signals the event loop invoked by fdb_run_network() to terminate. **You must call this function and wait for fdb_run_network() to return before allowing your program to exit, or else the behavior is undefined.**

Speedy37:
yep, this is the cause, `fdb_stop_network` **MUST** be called before the program exits.

Speedy37:
Internally I traced the segfault to a data race between the network thread that is still running and the `__do_global_dtors__aux` call in `libfdb_c.so`.

```
main thread:
    frame #22: 0x00007ffff7cb8ddf libfdb_c.so`TraceLog::~TraceLog(this=0x00007ffff7faa980) at Trace.cpp:496:14
    frame #23: 0x00007ffff74432de libc.so.6`__cxa_finalize + 206
    frame #24: 0x00007ffff77f5f47 libfdb_c.so`__do_global_dtors_aux + 39
    frame #25: 0x00007ffff7fe13fb ld-2.30.so`___lldb_unnamed_symbol59$$ld-2.30.so + 523
    frame #26: 0x00007ffff7442ba7 libc.so.6`___lldb_unnamed_symbol193$$libc.so.6 + 247
    frame #27: 0x00007ffff7442d60 libc.so.6`exit + 32
    frame #28: 0x00007ffff74201ea libc.so.6`__libc_start_main + 250
    frame #29: 0x000055555556c17e tokio-8ff57d870675dfc8`_start + 46
```

[...]

PR #177, opened 2020-02-24
Add NEGINSTART and POSINTEND to tuple

garrensmith:
Adds support for NEGINSTART and POSINTEND so that the tuple binding tests pass.

I'm still pretty new to Rust so the code might not be the perfect rust way of doing it. But it does make the binding layer tests for tuples pass which is cool 🎉


Speedy37:
Some related information about this, to help me review.

From https://github.com/apple/foundationdb/blob/master/design/tuple.md 

## Negative arbitrary-precision Integer

Typecodes: 0x0a, 0x0b Encoding: Not defined yet
Status: Reserved; 0x0b used in Python and Java

These typecodes are reserved for encoding integers larger than 8 bytes. Presumably the type code would be followed by some encoding of the length, followed by the big endian one’s complement number. Reserving two typecodes for each of positive and negative numbers is probably overkill, but until there’s a design in place we might as well not use them. In the Python and Java implementations, 0x0b stores negative numbers which are expressed with between 9 and 255 bytes. The first byte following the type code (0x0b) is a single byte expressing the number of bytes in the integer (with its bits flipped to preserve order), followed by that number of bytes representing the number in big endian order in one's complement.

## Positive arbitrary-precision Integer

Typecodes: 0x1d, 0x1e Encoding: Not defined yet
Status: Reserved; 0x1d used in Python and Java

These typecodes are reserved for encoding integers larger than 8 bytes. Presumably the type code would be followed by some encoding of the length, followed by the big endian one’s complement number. Reserving two typecodes for each of positive and negative numbers is probably overkill, but until there’s a design in place we might as well not use them. In the Python and Java implementations, 0x1d stores positive numbers which are expressed with between 9 and 255 bytes. The first byte following the type code (0x1d) is a single byte expressing the number of bytes in the integer, followed by that number of bytes representing the number in big endian order.

Speedy37:
Nice one, I didn't notice the tuple test only required this.
Do you have any idea why Github actions aren't running?

Also we can now add i128 and u128 support to the tuple layer.
And maybe "num-bigint" as an optional dependency.
Those changes aren't necessary to merge this.

Speedy37 (changes requested):

garrensmith:
Awesome. Thanks for reviewing so quickly. I'll make the changes tomorrow. No idea why the actions are not triggering. Is there anything I can do to trigger that?

Speedy37:
I guess they are not triggered because I only wrote `on: [push]` in `.github/workflows/ci.yml`.
I removed the `on: [push, pull_request]` to not double trigger the checks twice but it seems in this case it might be the cause.

garrensmith:
@Speedy37 thanks for the review. I've made all the required fixes.

garrensmith:
@Speedy37 could we for the checks to rerun again? I'm not sure why they are all failing

Speedy37:
The initial run failed due to #170 and when I pressed re-run checks it failed at checkout...
I've made a few more changes on my side that I will push soon (more tests, code reuse and i128, u128 support).

Speedy37:
Also I'm a bit confused by the current state of NEGINSTART and POSINTEND in the official repository:

This is wrong:
```cpp
if(data[i] >= NEG_INT_START && data[i] <= POS_INT_END) {
    i += abs(data[i] - INT_ZERO_CODE) + 1;
}
```
https://github.com/apple/foundationdb/blob/4f6054a3e6ff9321de3ff26b39b6151ada83e725/bindings/flow/Tuple.cpp#L108 

And in the java tuple layer tests, this seems wrong:

```
Tuple.from(BigInteger.valueOf(Long.MIN_VALUE).subtract(BigInteger.ONE)), new byte[]{0x0C, 0x7f, FF, FF, FF, FF, FF, FF, FF - 1},
```
https://github.com/apple/foundationdb/blob/b52eb746fdce92e4e0b44b96c54716836c44d518/bindings/java/src/test/com/apple/foundationdb/test/TupleTest.java#L121

Speedy37:
Nevermind, it does pass the test, it's because we only requires bigint encoding when below -2^64.

Speedy37 (approved):

garrensmith:
@Speedy37 some really nice improvements. Thanks for those

PR #184, opened 2020-03-09
fix use after free in Database::new

### LINKED ISSUE #183: Database::new doesn't pass fdb.cluster custom path
I noticed a problem trying to pass a non-standard path to fdb.cluster. I believe it's a matter of string getting deallocated after pointer is captured. The following patch seems to fix:

```
--- a/foundationdb/src/database.rs
+++ b/foundationdb/src/database.rs
@@ -46,10 +46,8 @@ impl Drop for Database {
 impl Database {
     /// Create a database for the given configuration path if any, or the default one.
     pub fn new(path: Option<&str>) -> FdbResult<Database> {
-        let path_str = path.map(|path| std::ffi::CString::new(path).unwrap());
-        let path_ptr = path_str
-            .map(|path| path.as_ptr())
-            .unwrap_or(std::ptr::null());
+        let path_str = std::ffi::CString::new(path.unwrap_or("")).unwrap();
+        let path_ptr = path_str.as_ptr();
         let mut v: *mut fdb_sys::FDBDatabase = std::ptr::null_mut();
         let err = unsafe { fdb_sys::fdb_create_database(path_ptr, &mut v) };
         error::eval(err)?;
```

Speedy37:
Resolves #183

### DIFF
--- a/foundationdb/src/cluster.rs
+++ b/foundationdb/src/cluster.rs
@@ -39,9 +39,12 @@ impl Cluster {
     ) -> impl Future<Output = FdbResult<Cluster>> + Send + Sync + Unpin {
         let path_str = path.map(|path| std::ffi::CString::new(path).unwrap());
         let path_ptr = path_str
+            .as_ref()
             .map(|path| path.as_ptr())
             .unwrap_or(std::ptr::null());
-        FdbFuture::new(unsafe { fdb_sys::fdb_create_cluster(path_ptr) })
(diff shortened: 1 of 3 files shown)


Speedy37:
The fix could have been limited to only add `as_ref` in order to not consume path_str.

Speedy37:
Tests catched the same error in the deprecated (fdb < 6.0) cluster api.

PR #186, opened 2020-03-12
[WIP] Directory

garrensmith:
This is the PR to get the directory layer working with the rust client. Its currently still very rough
and doesn't even compile yet.

If any of the maintainers want to contribute feel free to push code. I'm also on the rust discord (redcomet) if you want to chat about it.

### DIFF
--- a/foundationdb/src/directory/node.rs
+++ b/foundationdb/src/directory/node.rs
@@ -1,0 +1,65 @@
+use crate::tuple::{Subspace};
+use crate::future::FdbSlice;
+use crate::{Transaction, Directory, DirectoryResult};
+use crate::directory::directory_subspace::DirectorySubspaceResult;
+
+
+pub struct Node {
+    subspace: Subspace,
+    path: Vec<String>,
+    target_path: Vec<String>,
+    _layer: Option<Vec<u8>>
+}
+
+impl Node {
+
+    pub fn exists(&self) -> bool {
+        // if self.subspace == None {
+        //     return false;
+        // }
+
+        true
+    }
+
+    pub async fn prefetchMetadata(&self, trx: &Transaction) -> &Node {
+        if self.exists() {
+           self.layer(trx).await;
+        }
+
+        return self;
+    }
+
+    pub async fn layer(&mut self, trx: &Transaction) -> &[u8] {
+        if self._layer == None {
+            let key = self.subspace.subspace(&b"layer".to_vec());
+            self._layer = match trx.get(key.bytes(), false).await {
+                Ok(None) => Some(Vec::new()),
+                Err(_) => Some(Vec::new()),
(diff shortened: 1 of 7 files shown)


garrensmith:
Just to add I've based the initial work off the golang and python implementation. I've also used the [notes](https://github.com/cloudant-labs/couchdb-erlfdb/blob/master/notes/directory-layer-design.md) from the erlang implementation that we using for CouchDB.

scripts/run_bindingtester.sh
```diff
@@ -6,7 +6,7 @@ fdb_rs_dir=$(pwd)
 
 case $(uname) in
   Darwin)
-    brew install mono
+#    brew install mono
```
Speedy37:
```suggestion
command -v mono || brew install mono
```

scripts/run_bindingtester.sh
```diff
@@ -22,7 +22,7 @@ esac
   cd ${fdb_builddir:?}
 
   ## Get foundationdb source
-  git clone --depth 1 https://github.com/apple/foundationdb.git -b release-6.1
+#  git clone --depth 1 https://github.com/apple/foundationdb.git -b release-6.1
```
Speedy37:
```suggestion
[[ -d foundationdb ]] || git clone --depth 1 https://github.com/apple/foundationdb.git -b release-6.1
```

foundationdb/src/directory/node.rs
```diff
@@ -0,0 +1,65 @@
...
+
+        return self;
+    }
+
+    pub async fn layer(&mut self, trx: &Transaction) -> &[u8] {
+        if self._layer == None {
+            let key = self.subspace.subspace(&b"layer".to_vec());
+            self._layer = match trx.get(key.bytes(), false).await {
+                Ok(None) => Some(Vec::new()),
+                Err(_) => Some(Vec::new()),
```
Speedy37: I don't think silencing an FdbError is ok.
layer should probably return a Result

foundationdb/src/directory/node.rs
```diff
@@ -0,0 +1,65 @@
...
+use crate::future::FdbSlice;
+use crate::{Transaction, Directory, DirectoryResult};
+use crate::directory::directory_subspace::DirectorySubspaceResult;
+
+
+pub struct Node {
+    subspace: Subspace,
+    path: Vec<String>,
+    target_path: Vec<String>,
+    _layer: Option<Vec<u8>>
```
Speedy37: Layer being used as a cache you might want to consider https://docs.rs/futures/0.3.4/futures/future/trait.FutureExt.html#method.shared.
RefCell might also be required in order to not requires `&mut self` on a lot of methods (but be careful about dynamic borrows lifespans in async context).

PR #194, opened 2020-06-27
Add support for pack with versionstamp

### LINKED ISSUE #192: Missing offset for versionstamp
When using `atomic_op` with `MutationType::SetVersionstampedKey`, the key should have the offset of the versionstamp position.

E.g
```rust
let key = pack(&("foo", Versionstamp::incomplete(0)))
// [2, 102, 111, 111, 0, 51, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 0, 0]

fdb.atomic_op(key, "bar".as_ref(), MutationType::SetVersionstampedKey).await?;
// throws: thread 'main' panicked at 'failed to run: FdbError { error_code: 2000 }', src/main.rs:32:9
```

Expected behavior:
```rust
let key = pack_with_versionstamp(&("foo", Versionstamp::incomplete(0)))
// [2, 102, 111, 111, 0, 51, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 0, 0, 6, 0, 0, 0]

fdb.atomic_op(key, "bar".as_ref(), MutationType::SetVersionstampedKey).await?;
// 👍 
```

### LINKED ISSUE #193: Typo in bindingtester
`TUPKE_PACK_WITH_VERSONSTAMP` should probaby be `TUPLE_PACK_WITH_VERSIONSTAMP`

https://github.com/Clikengo/foundationdb-rs/blob/94b05f34f50a653734b53fff29078003b75fd4f8/foundationdb-bindingtester/src/main.rs#L236

### LINKED ISSUE #195: Run bindingtester with the same configuration as python
python is running the bindingtester with different values: https://github.com/apple/foundationdb/blob/2105778026f17cb1bf186e3b502327d49f6e92ce/bindings/bindingtester/known_testers.py#L60

Running with the same setting as python should test both the versionstamp api and the bigint support

Speedy37:
Doesn't change the packing format.
Fixes a bug in the unpacking of signed integers.
Add support for BigInt on Element
Element now implement Ord and matches the packed ordering.
Add support for packing with versionstamp
Enhance testing to match python output.

Resolves #192
Resolves #193
Resolves #195

### DIFF
--- a/foundationdb/src/tuple/element.rs
+++ b/foundationdb/src/tuple/element.rs
@@ -3,7 +3,9 @@ use super::{Bytes, Versionstamp};
 use std::{borrow::Cow, cmp};
 
 #[cfg(feature = "num-bigint")]
-use std::convert::TryInto;
+use num_bigint::Sign;
+#[cfg(feature = "num-bigint")]
+use std::convert::TryFrom;
 
 #[derive(Clone, Debug)]
 pub enum Element<'a> {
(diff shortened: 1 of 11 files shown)


foundationdb/src/tuple/element.rs
```diff
@@ -30,6 +32,8 @@ impl<'a> Element<'a> {
             Element::Tuple(v) => Element::Tuple(v.into_iter().map(|e| e.into_owned()).collect()),
             #[cfg(feature = "uuid")]
             Element::Uuid(v) => Element::Uuid(v),
+            #[cfg(feature = "num-bigint")]
+            Element::BigInt(v) => Element::BigInt(v),
```
garrensmith, line 36: Could you add a `as_versionstamp` conversion below. That would be really useful.

Speedy37: ok will do.

Speedy37:
This is starting to become quite a big commit. I've got massive changes to the bindingtester so it behave exactly like python.

garrensmith:
Do you want to split it into two?

________________________________

Speedy37:
I still need to find out why get_range sometimes return a 2000 error code.

Speedy37:
And I had to use unsafe to bypass rust protection against bad usage of cancelled transaction.

garrensmith:
This is awesome. Thank you

PR #196, opened 2020-07-11
fix broken transact() limits

dae:
The logic was backwards - if a retry limit is set (say 50), 1 < 50 is
true, so is_none() is false and transact() immediately aborts.


Speedy37:
good catch

Speedy37:
I was unable to push more commit into your branch, so I had to create another PR #197

dae:
Thanks!

Also a thank you for the work on the 0.4 branch - I recently got around to updating my code from an async/await branch of 0.3, and the API in 0.4 is considerably nicer.

PR #214, opened 2020-12-18
make tuple::hca::HighContentionAllocator be Send

zerth:
- thread rng is not `Send`; instead, seed a `SmallRng` from it, which is `Send`
- use of allocation mutex was causing non-`Send`ability despite being dropped before any `.await` boundaries; wrapped in blocks
- update related test to use `.boxed()` instead of `.boxed_local()`

### DIFF
--- a/foundationdb/Cargo.toml
+++ b/foundationdb/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "foundationdb"
-version = "0.5.2"
+version = "0.5.1"
 authors = [
     "Benjamin Fry <benjaminfry@me.com>",
     "Vincent Rouillé <vincent@clikengo.com>",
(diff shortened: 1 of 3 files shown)


foundationdb/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "foundationdb"
-version = "0.5.1"
+version = "0.5.2"
```
Speedy37, line 3: Can you remove this version bump?

Speedy37 (changes requested):
Looks good overall. CI is broken tho.

foundationdb/Cargo.toml
zerth: Done!

PR #217, opened 2020-12-31
Directory layer

### LINKED ISSUE #27: Higher-layer API: directory/subspace/tuple
In addition to key-value API, there are some API layers which are supported by official bindings and compatible between them (Python/Go/Java).

For Java: https://apple.github.io/foundationdb/javadoc/index.html
For Golang
 - https://godoc.org/github.com/apple/foundationdb/bindings/go/src/fdb/directory
 - https://godoc.org/github.com/apple/foundationdb/bindings/go/src/fdb/subspace
 - https://godoc.org/github.com/apple/foundationdb/bindings/go/src/fdb/tuple

It's great to have those layers in Rust.

PierreZ:
This is the PR draft to resolves #27 and based on the initial work done by @garrensmith available on #186.

I wrote some basic tests, but here's a TODO list:

- [x] Add more tests on the `Node` part,
- [x] make clippy happy,
- [x] Add `layer` verification on the last node before opening,
- [x] Implement `list`,
- [x] Implements `move`,
- [x] Implements `delete`,
- [x] Add directory to [BindingTester](https://github.com/apple/foundationdb/blob/master/bindings/bindingtester/spec/directoryLayerTester.md)
- [x] ~~Add validation that the binding is  compatible with others~~
- [x] Implement directory partitions
- [x] check ranges
- [x] Passing all seeds and steps
- [x] cleanup code + doc

My rust is feeling a bit rusty, so feel free to comment :smile: 

EDIT: I discovered bindingTester while searching for directory validation.


foundationdb/src/directory/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+            Some(node) => node,
+        };
+        let content_subspace = last_node_from_old_path
+            .content_subspace
+            .as_ref()
+            .ok_or(DirectoryError::DirNotExists)?;
+
+        let mut new_nodes = self.find_nodes(&trx, new_path.to_owned()).await?;
+        // assert that parent of the new node exists
+        if new_nodes.get(new_nodes.len() - 2).is_none() {
```
Speedy37: panic if `len() < 2`

foundationdb/src/directory/node.rs
```diff
@@ -0,0 +1,120 @@
...
+        let mut results = vec![];
+
+        let range_option = RangeOption::from(&self.node_subspace.to_owned());
+
+        let fdb_values = trx.get_range(&range_option, 1_024, false).await?;
+
+        for fdb_value in fdb_values {
+            let subspace = Subspace::from_bytes(fdb_value.key());
+            // stripping from subspace
+            let sub_directory: Vec<u8> = self.node_subspace.unpack(subspace.bytes())?;
```
Speedy37: You unpack bytes to bytes? hum...
What does fdb_value.key() content is?
Why not unpacking the string directly?

Cpp code does: `subdir.unpack(subdirRange[i].key).getString(0)`
Which in rust should be equivalent to `self.node_subspace.unpack::<(String,)>(key)?`

Speedy37 (commented):
I did a quick review of the code. Looks like you are making good progress 😀

foundationdb/src/directory/mod.rs
PierreZ: You are right, I backported the code from the Go binding, and I'm wondering why we cannot move on the root-node. I'll definitely add more restrictions :smile:

foundationdb/src/directory/node.rs
PierreZ: > Why not unpacking the string directly?

That is a good question indeed :thinking: I will change this, thanks!

EDIT: after testing, I know why I did this:  got a ` PackError(BadCode { found: 1, expected: Some(2) })` when trying to unpack to (String,). I think `1` stands for the [const BYTES: u8 = 0x01;](https://github.com/Clikengo/foundationdb-rs/blob/master/foundationdb/src/tuple/mod.rs#L26) so I must have stored everything in bytes. The keys are generated with `subspace.pack(something_in_string.bytes())`, and I did not used the turbofish syntax (love the name :rofl:), because I was crafting key's bytes by hand, like I'm used to do in HBase.

I will make some changes and properly create keys that can be easily deserialize. Go's binding is doing [this](https://github.com/apple/foundationdb/blob/master/bindings/go/src/fdb/directory/directoryLayer.go#L168):

```go
tr.Set(parentNode.Sub(_SUBDIRS, path[len(path)-1]), prefix)
```

And [Sub](https://pkg.go.dev/github.com/apple/foundationdb/bindings/go/src/fdb/subspace#Sub) is accepting a TupleElement, so I should be OK regarding binding's compat :sunglasses: 

Thanks a lot :+1:

PierreZ:
Thanks for the early review @Speedy37 :+1:

PierreZ:
I'm having difficulties to grasp the idea of the `DirectoryPartition`, do you have knowledge of it, or should I ask on the forum?

PierreZ:
Turns out, my first implementation and mental representation of the `Directory` was wrong :stuck_out_tongue: 

Previous commits were not compatible with other bindings, so everything has been manually checked with the `Go` binding. Before reimplement `delete` and `move`, I will include the `binding-checker` to validate that everything is working properly.

Speedy37:
Great work :)

For the bindingtester, you can add the test to https://github.com/Clikengo/foundationdb-rs/blob/master/scripts/run_bindingtester.sh so the CI runs it.

test-name is directory I think

PierreZ:
Just wanted to share that I'm finally passing a seed :partying_face: 

```
Test with seed 1330929912 and concurrency 1 had 0 incorrect result(s) and 0 error(s) at API version 610
Completed directory test with random seed 1330929912 and 10 operations
```

TODO list updated

garrensmith:
Wow awesome work.

________________________________

PierreZ:
Hey :wave: 

Last week, as I was digging into some issues on my implementation, I had the idea to go through Flow´s implementation(to be honest, I was only looking at the Go/Java´s code :tongue:) After discovering that Rust async is closer to Flow than I thought, I decided to erase everything and just backport the Flow´s implementation, which will be more maintainable in my opinion :smile: 

However, I stumble across two issues related to async and `FdbValuesIter` which is not `Send`. The related code can be found  [here](https://github.com/PierreZ/foundationdb-rs/blob/dev/directory/foundationdb/src/directory/directory_layer.rs#L665-L706):

```
error: future cannot be sent between threads safely
   --> foundationdb/src/directory/directory_layer.rs:665:5
    |
665 |     #[async_recursion]
    |     ^^^^^^^^^^^^^^^^^^ future created by async block is not `Send`
    |
    = help: within `FdbValuesIter`, the trait `std::marker::Send` is not implemented for `Rc<FdbFutureHandle>`
note: future is not `Send` as this value is used across an await
   --> foundationdb/src/directory/directory_layer.rs:688:17
    |
679 |             let range: Arc<FairMutex<FdbValuesIter>> = Arc::new(FairMutex::new(range.into_iter()));
    |                 ----- has type `Arc<parking_lot::lock_api::Mutex<RawFairMutex, FdbValuesIter>>` which is not `Send`
...
688 |                 self.remove_recursive(trx, sub_node).await?;
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ await occurs here, with `range` maybe used later
...
694 |         }
    |         - `range` is later dropped here
    = note: required for the cast to the object type `dyn futures::Future<Output = std::result::Result<(), DirectoryError>> + std::marker::Send`
    = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
```
error: future cannot be sent between threads safely
   --> foundationdb/src/directory/directory_layer.rs:665:5
    |
665 |     #[async_recursion]
    |     ^^^^^^^^^^^^^^^^^^ future created by async block is not `Send`
    |
    = help: within `FdbValuesIter`, the trait `std::marker::Send` is not implemented for `*const keyvalue`
note: future is not `Send` as this value is used across an await
   --> foundationdb/src/directory/directory_layer.rs:688:17
    |
679 |             let range: Arc<FairMutex<FdbValuesIter>> = Arc::new(FairMutex::new(range.into_iter()));
    |                 ----- has type `Arc<parking_lot::lock_api::Mutex<RawFairMutex, FdbValuesIter>>` which is not `Send`
...
688 |                 self.remove_recursive(trx, sub_node).await?;
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ await occurs here, with `range` maybe used later
...
694 |         }
    |         - `range` is later dropped here
    = note: required for the cast to the object type `dyn futures::Future<Output = std::result::Result<(), DirectoryError>> + std::marker::Send`
    = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
`FdbValuesIter` is defined like this:

```rust
/// An iterator of keyvalues owned by a foundationDB future
pub struct FdbValuesIter {
    f: Rc<FdbFutureHandle>,
    keyvalues: *const fdb_sys::FDBKeyValue,
    len: i32,
    pos: i32,
}
 ```

and the first two fields are the one not `Send `able. Could you help me fix it? I tried the Arc<Mutex<>> technique, without luck. Thanks in advance :smiley:

Speedy37:
Hi,

Nice work 👏

About the Send issue, I see 2 ways to try to fix it :

 1. Very simple: maybe `FdbValuesIter` and `FdbValue` should contains an `Arc<FdbFutureHandle>`. Then you are allowed to add `unsafe impl Send for FdbValuesIter {}` and `unsafe impl Send for FdbValue {}`. The lifetime of `*const fdb_sys::FDBKeyValue` is the same as the handle.
 2. Return a non Send future: don't use `#[async_recursive attribute]` and write the workaround yourself (wrapper method that box the future to have a finite stack size).

I would personally go for 1 I think.

PierreZ:
> Hi,
> 
> Nice work clap
> 
> About the Send issue, I see 2 ways to try to fix it :
> 
>     1. Very simple: maybe `FdbValuesIter` and `FdbValue` should contains an `Arc<FdbFutureHandle>`. Then you are allowed to add `unsafe impl Send for FdbValuesIter {}` and `unsafe impl Send for FdbValue {}`. The lifetime of `*const fdb_sys::FDBKeyValue` is the same as the handle.
> 
>     2. Return a non Send future: don't use `#[async_recursive attribute]` and write the workaround yourself (wrapper method that box the future to have a finite stack size).
> 
> 
> I would personally go for 1 I think.

Thanks a lot @Speedy37, option 1 worked perfectly :rocket:

PierreZ:
I finished my first iteration of backporting Flow´s code, now I´m working on making tests passes :smile:

PierreZ:
Most of my tests are passing on most seeds, except when I'm using the `DirectoryPartitions`. I'm struggling to find the problem here for now, keep digging

PierreZ:
Still digging :disappointed: I am experiencing quite some difficulties fixing the last bugs, as fixing one seed is creating new bugs in seeds that used to pass :tongue: 

Most of my mistakes seems to be related to:

* layer (sometimes I cannot find the layer under the right key). I tried removing every `clear` from my implementation, without luck
* `DirectoryPartition` (`exists` and `list` are failing most of the time), due to path's manipulation between the directory Partition/Subspace/Layer. I'm backporting [this class from Flow/C++](https://github.com/apple/foundationdb/blob/master/bindings/flow/DirectoryPartition.h) but I must admit my C++ is even more rusty than my rust :smiley:  I'm not sure if the DirectoryPartition is used a lot, but we cannot disable it from the bindingtester.

I could use some help :innocent: If someone is willing to help me, I've marked some faulty seeds [here](https://github.com/PierreZ/foundationdb-rs/blob/0d71d80882c1b1b555ac4d757315162666aad358/foundationdb-bindingtester/README.md#directory-faulty-seed).

PierreZ:
Unblocked, thanks to @Geal :tada: 

Got a dedicated computer searching for faulty seed with 10k ops. I will fix them and cleanup the code :smile:

PierreZ:
The bindingtester have been running for 24h+ with 10k ops without finding a faulty seed :smile: 

I squashed the commits, ready to be reviewed :tada: 

As mentionned before, I tried to be as close as possible from the Flow code, but it can be changed.

pierrebelzile:
I was looking for this functionality today. I see that was a large amount of work. Has been in review for nearly a year. What would be next step for the merge?

PierreZ:
> I was looking for this functionality today. I see that was a large amount of work. Has been in review for nearly a year. What would be next step for the merge?

Hi @pierrebelzile :wave: 

Please note that the development of the crate has moved to the dedicated [Github org](https://github.com/foundationdb-rs/foundationdb-rs). The fork is including the directory feature, and several others. Support for FDB 7.1 is also on its way.

More info can be found here on why we had to hard fork can be found [here](https://github.com/foundationdb-rs/foundationdb-rs/issues/1).

pierrebelzile:
Thanks. Can't believe I missed that. Great news and thanks for the work! :)

PR #233, opened 2021-07-09
ci: remove windows

PierreZ:
The FDB community is officially discontinuing the Windows build (see
apple/foundationdb#5135 for more info). We cannot download msi packages
as they were removed from official sites.

As the `windows-latest` is mandatory for any PR, this commit is removing
the `windows-latest` from Github actions. We may add it back later to
use community build.


PierreZ:
Mmmh, I can see some errors related to [clap](https://github.com/PierreZ/foundationdb-rs/runs/3336800152?check_suite_focus=true#step:5:152). [Clap's MSRV for >=3.0 is 1.54.0](https://github.com/clap-rs/clap#minimum-supported-version-of-rust-msrv), should we bump? 

Also, we are targetting `structopt = "0.3.3"` which does not exists, latest is [0.3.22](https://crates.io/crates/structopt/versions?sort=semver).

Speedy37:
bindingtester is only "in tree" so we probably don't have to care about MSRV.

Speedy37:
but CI has (in the current configuration) to run with 1.40...

PierreZ:
CI is [green on my side](https://github.com/PierreZ/foundationdb-rs/actions/runs/1277167912), but I had to bump the MSRV to fix it, `clang-sys`'s transitives dependencies requires 1.46.

